PR #11, opened 2023-02-07
Fix python requirements for numpy

a-komarov-playrix:
*Issue #, if available:*
`python -m pip install -r requirements.txt`
Pip raises error:
`ERROR: No matching distribution found for numpy==1.23.1`

*Description of changes:*
Fixed possible typo.


switch180:
Hi! Thanks for your PR.

I believe the issue is that [numpy 1.23.1](https://pypi.org/project/numpy/1.23.1/) is only available for Python >=3.8.

Are you using Python 3.7? [v1.21.3 is available for this version of Python](https://pypi.org/project/numpy/1.21.3/).

I think instead of modifying the requirements, we should add a badge and/or explanation to the README explaining we require Python version 3.8 or higher (the pytest dependency requires v3.7 or higher btw).  

Could you keep the requirements the same, but make changes to the README etc? It looks like shields.io is the standard for badges; we might need to make some changes to the top level folder to make these work https://github.com/badges/shields

No worries if you can't

a-komarov-playrix:
You're right, I'm using python 3.7 and I overlooked recommendation to use 3.9 in readme.
From this point I think nothing ought to be done. Thank you for your time!

PR #23, opened 2023-09-14
feat: Added profile parameter

gonzariosm:
*Changes:*

Added the `--profile` option to pass a different profile if you manage multiple AWS accounts.


jakob-vendegna-sp:
FYI I filed a support case about this with AWS. It seems the maintainer of this repo is out on an extended leave with no defined return date. So... I guess this is on-hold until further notice. 😢

jakob-vendegna-sp:
Oh, but it wouldn't hurt to purge those pycache files in your PR in the meantime. I know they'll ask about that first.

capacity-mode-evaluator/src/pricing.py
```diff
@@ -5,7 +5,7 @@
 
 
 class PricingUtility(object):
-    def __init__(self, region_name):
+    def __init__(self, region_name, profile_name='default'):
```
tebanieo, line 8: This change will force everyone to use a profile, there are customers that use this tool with environment variables.

capacity-mode-evaluator/src/pricing.py
```diff
@@ -17,8 +17,9 @@ def __init__(self, region_name):
         # pick the closest endpoint to the supplied region
         if region_name not in AMERICAN_REGIONS:
             closest_api_region = 'ap-south-1'
-
-        self.pricing_client = boto3.client(
+        
+        self.session = boto3.session.Session(profile_name=profile_name)
```
tebanieo, line 21: Implement the necessary conditional and modify the code to accommodate None as an input for customers utilizing environment variables.

table_tagger.py
```diff
@@ -88,6 +88,9 @@ def main():
 
     parser.add_argument(
         '--tag-name', required=False, type=str, default='table_name',  help='tag table with tag TAG_NAME (default is "table_name")')
+    
+    parser.add_argument(
+        '--profile', required=False, type=str, default='default', help='set a custom profile name to perform the operation under')
```
tebanieo, line 93: The default value should be none, what if you want to use the environment variables instead of the client?

tebanieo (changes requested):
Hello Thanks for the PR, merging this code will add a strong dependency on having set up your profile, there are some situations where customers work with environment variables instead of the profile file. Please make this optional as we don't want to force anyone to use the default `.profile` file.

ddbtools/__pycache__/constants.cpython-311.pyc
tebanieo, line 1: add `*.pyc` to your .git ignore so you don't push the pycache.

capacity-mode-evaluator/src/pricing.py
gonzariosm: Updated to `None`

capacity-mode-evaluator/src/pricing.py
gonzariosm: Updated to `None`

table_tagger.py
gonzariosm: Updated to `None`

ddbtools/__pycache__/constants.cpython-311.pyc
gonzariosm: Added some lines in `.gitignore`and removed `__pycache__`

gonzariosm:
@tebanieo can you take a look and let me know if I can add something else to this PR, please?

tebanieo (approved):

PR #38, opened 2024-07-23
Update config.py (TLV Region)

irmitz:
Add support to TLV (il-central-1) region, including pricing.

*Issue #, if available:*
The script failed if any capacity/dto consumed in TLV region ("MEC1-")

*Description of changes:*
Added TLV region to the config.py to fix the issue.


switch180:
Looks good to me!

PR #42, opened 2024-11-14
Updating rWCU prices for provisioned mode

switch180:
*Issue #, if available:*

*Description of changes:*
We don't use this pricing data, but we have it so it should be up to date.


switch180:
Fixed in #44

PR #140, opened 2026-04-23
Adds the new "load-export" verb

ruskindantra:
*Description of changes:*
This new verb (`load-export`) allows you to load (import) DynamoDB full or incremental exports. You can do so into a clean table or one with data. It will overwrite any existing data if the keys match. The verb is also augmented with a `transform` capability which let's you manipulate the data and the action before the items are written into the table. There are some example of what common transforms might look like. The code is modular with full unit tests supporting.

*Usage*:
```
./bulk load-export --table <target> --s3-path <s3://bucket/path/to/data> [--transform <transform_module>]
```


tools/bulk_executor/client/src/infrastructure/bootstrap.py
```diff
@@ -195,6 +195,8 @@ def _create_or_update_glue_job(self, args, is_create_allowed=True):
 
         log4j2_properties_file_location = f's3://{glue_job_bucket}/{LOG4J_PROPERTIES_FILE}'
 
+        enable_continuous_cloudwatch_log = bool(args.get('XContinuousLogging'))
```
hunterhacker, line 198: I don't recall having a discussion about this. Can you pls give me background?

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/__init__.py
```diff
@@ -54,13 +56,18 @@ class RateLimiterWorker:
     Args:
         shared_config (RateLimiterSharedConfig): The shared config between Aggregator and Worker.
         monitor_options: The expected monitor options (see @table_info#get_dynamodb_throughput_configs for more info)
+        debug_accumulator (DebugAccumulator): The debug accumulator to use for rate limiter
```
hunterhacker: Why do we need a debug accumulator? And is there a risk it grows too big on a long run?

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
```diff
@@ -41,6 +43,18 @@ def __init__(self, session, bucket, prefix,
         self.worker_max_read_rate = worker_max_read_rate
         self.worker_max_write_rate = worker_max_write_rate
 
+        if self.debug_accumulator:
+            self.debug_accumulator.add([f"DistributedDynamoDBMonitorWorker init: worker_id={self.worker_id}"])
```
hunterhacker: Doesn't this wait til the end of execution to be gathered? You don't want it available during execution? What's the philosophy behind this?

tools/bulk_executor/server/src/python_modules/shared/table_info.py
```diff
@@ -57,7 +55,7 @@ def get_quota_value(quota_name, region_name):
         log.info(f"Warning: Could not retrieve quota: {str(e)}")
         return None
 
-def get_and_print_dynamodb_table_info(table_name, index_name=None):
+def get_and_print_dynamodb_table_info(table_name, index_name=None, quiet=False):
```
hunterhacker, line 58: I always felt like get_and_print was a bit weird. Now with a quiet param, it's get_and_print with a flag to make it just get. Extra weird now, right.

tools/bulk_executor/server/src/root.py
```diff
@@ -62,6 +63,11 @@ def _get_parsed_glue_job_args(argv):
 parsed_args = _get_parsed_glue_job_args(sys.argv)
 
 action_module = parsed_args.get('XAction', 'default')
+
+# Handle hyphen-to-underscore mapping for Python module names
```
hunterhacker: No hard coding in this file. Extensibility should be available by just putting the right python code in the right place, not hand-editing root.py. Please just convert hyphens to underscores.

tools/bulk_executor/bulk
```diff
@@ -79,6 +79,10 @@ if(action_input.XAction is None):
 
 action = vars(action_input)["XAction"] # XAction is always a required input.
 
+# Handle hyphen-to-underscore mapping for Python module names
+if action == 'load-export':
```
hunterhacker: No hard-coding of command names here either. Just convert hyphens in input to underscores for lookup.

tools/bulk_executor/README.md
```diff
@@ -109,6 +109,9 @@ Here are some example use cases:
 # If going cross-account, you need a resource-based policy on the table to allow access.
 ./bulk copy --source source --target target
 ./bulk copy --source arn:aws:dynamodb:us-east-1:123456789012:table/source --target arn:aws:dynamodb:us-west-2:987654321098:table/target
+
+# Load a DynamoDB export into an existing DynamoDB table
+./bulk load-export --table target --s3-path "s3://..." [--transform example] [--transformfunctionname transform]
```
hunterhacker: No more transform function name, right?

tools/bulk_executor/README.md
```diff
@@ -504,6 +507,20 @@ If doing cross-account, you need a resource-based policy to enable access. The f
 }
 ```
 
+#### `load-export`**
+
+* Leverage Glue to do a full/incremental load of a DynamoDB table which was exported to S3 (i.e. currently only supports DDB-JSON format)
```
hunterhacker: Everything here leverages Glue, prob don't need to mention that.

tools/bulk_executor/README.md
```diff
@@ -504,6 +507,20 @@ If doing cross-account, you need a resource-based policy to enable access. The f
...
+* Requires `table` and `s3-path`
+* Validates manifest files to ensure integrity of the exported files and data
+* Supports loading a full export into an empty table or table with data (will overwrite existing data if keys match)
+* Supports loading an incremental export, will add new data and only update/delete data if keys match
+* Best used with:
+  * `--XMaxWriteRate N` where _N_ is the max aggregate WCU to consume on the destination table
```
hunterhacker: Technically WRU/WCU since WCU is only for provisioned mode. I often say "write units" to avoid the ugliness while staying accurate.

tools/bulk_executor/README.md
```diff
@@ -504,6 +507,20 @@ If doing cross-account, you need a resource-based policy to enable access. The f
...
+  * `--transform` and `--transformfunctionname` where you specify the python function used to transform items during load
```
hunterhacker: Are we doing a function name?

tools/bulk_executor/README.md
```diff
@@ -514,9 +531,10 @@ A few parameters, each starting with `X` to differentiate them from regular scri
     * FLEX run: `Job duration: 0:11:34.849592 (7.57 DPU hours)`
 * `XNumberOfWorkers`: Controls how many Glue workers to use, default 220. Small tables or jobs could set this lower.
 * `XWorkerType`: Controls the worker type. Can be `G.1X` (default), `G.2X`, `G.4X`, `G.8X`, `G.12X`, `G.16X`, `R.1X`, `R.2X`, `R.4X`, or `R.8X`. Using a larger type provides more compute, memory, and storage for jobs that require a single worker to gather data locally, for example an `orderby` applied to a large set of matching items.
-* `XRetries`: Controls the max number of Glue Job retries. Defaults to zero to fail a misconfigured Glue Job quickly. It's unlikely you'll need to need to change this.
+* `XRetries`: Controls the max number of Glue Job retries. Defaults to zero to fail a misconfigured Glue Job quickly. It's unlikely you'll need to change this.
 * `XTimeout`: Controls the Glue Job timeout (in minutes). Default is 60 minutes.
 * `XWaitForDPU`: If present, causes the execution to wait 40 seconds at the end of execution in order for the DPU metrics to arrive so they can be reported. Default is off.
+* `XContinuousLogging`: If present, default Glue logging is used rather than Log4J
```
hunterhacker: I don't think any user will know what that means. I don't. :)

tools/bulk_executor/client/src/infrastructure/bootstrap.py
ruskindantra: During initial testing, I needed deep internal Glue exceptions which weren't available. As soon as you pass in the log4j file, it disables detailed logging. So this flag allows you to control that if detailed logging is ever required, which was the case when I was getting the setup working (folder structures etc)

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/__init__.py
ruskindantra: To help with debugging, it was needed to debug some peculiar scenarios with rate limiting within the context of loading the exports across multiple workers. Yes, there is a risk but it's a debug utility so if you turn it on, chance are you are solving a problem (the debug accumulator growing won't be one of those). Unless, you leave it "on" by accident.

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
ruskindantra: To understand whether the rate limiters were initialized correctly. Yes, it will only surface right at the end of the execution but that's okay.

tools/bulk_executor/server/src/python_modules/shared/table_info.py
ruskindantra: Yes, agreed, we should break this down into two, 1/ get call, 2/ print call. If we do that...then the quiet flag quietly goes away :)

I refrained from that chance as this function is used in a lot of different verbs and regression testing those will take time. We can create an issue around this.

tools/bulk_executor/README.md
ruskindantra: No, remnants of previous path, removed

tools/bulk_executor/README.md
ruskindantra: https://docs.aws.amazon.com/glue/latest/dg/monitor-continuous-logging-enable.html

It's for advanced users, allows you to get detailed logging out without touching code

tools/bulk_executor/client/src/infrastructure/bootstrap.py
hunterhacker: Why does asking for continuous logging cause the side effect of giving you detailed logging? 

Does it basically override and undo the log4j2.properties setting? 

Is this flag going to be an undocumented thing that you just know is handy sometimes?

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
```diff
@@ -18,7 +18,8 @@ def __init__(self, session, bucket, prefix,
                  worker_id=None,
                  summary_key='summary.json',
                  enable_reporting=False,
-                 autostart=True):
+                 autostart=True,
+                 debug_accumulator=None):
```
hunterhacker: I've tried to keep rate_limiter agnostic about its use in Bulk so someone could use it for other purposes. Even wrote a blog encouraging that use. Eventually it could be its own pip. I don't think we should pass a Glue-specific parameter in to it with a behavior that's very Bulk oriented.

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
hunterhacker: Why did you decide to accumulate logs for at-the-end printing instead of having the text go to CloudWatch right away and having a way to print that out via LiveTail?

tools/bulk_executor/server/src/python_modules/shared/table_info.py
hunterhacker: Yep. The functionality has accumulated and it was messy before and with this param it's ultra messy.

For the moment, to make it a little less messy today, what about this: have the function return two values (a) the data and (b) a string that could be printed if the caller wanted. And then the caller code can do the print or not. That removes the parameter and the vast number of "if" statements.

tools/bulk_executor/README.md
hunterhacker: Yep, read that page, remain confused. :) 

We had continuous logging even with this set false, did we not? I mean, data was showing up in the logs without perceivable delay, right? 

And somehow asking for continuous logging has the side effect of getting detailed logging. "Oh you want detailed logging, you turn on continuous logging." How's that a thing?

tools/bulk_executor/client/src/infrastructure/bootstrap.py
ruskindantra: Yes, that's the behavior, I have added a reference on line 221 - https://repost.aws/knowledge-center/glue-reduce-cloudwatch-logs.

We can provide documentation but given it's a deep capability I would expect 99% of the time that we don't turn it on

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
ruskindantra: That's a good point, yes we did talk about extracting this out into a npm/pip module (already done as part of C# as a POC)

tools/bulk_executor/client/src/python_modules/load_export.py
```diff
@@ -0,0 +1,43 @@
+import argparse
+import logging as log
+
+import utils
+from utils.custom_parser import BulkArgumentParser
+
+help_text = f"""
+    Purpose of "load-export":
+        Loads a full export from S3 to an existing DynamoDB table.
+        Loads a full export from S3 to an existing DynamoDB table.
```
hunterhacker, line 9: Got the same line twice :)

tools/bulk_executor/README.md
hunterhacker: I propose we don't document this publicly. It's a developer thing which if you read the code you can learn about it. The docs seem to imply that in Glue 5 it won't even be relevant anymore, and we'll prob want to move to Glue 5 soon as we can.

tools/bulk_executor/client/src/python_modules/fill.py
```diff
@@ -33,6 +33,8 @@ def run(env_configs):
 
     utils.validate_tables(env_configs, parser, result['table'], pitr_enabled=True)
 
+    result['generator'] = utils.sanitize_arg(result['generator'], r'\.py$')
```
hunterhacker: Would be good to add a comment here that we want to allow the .py suffix if someone does it by accident. Same in the other verbs with this trick.

tools/bulk_executor/client/src/runner.py
```diff
@@ -104,6 +105,12 @@ def _pretty_print_log_event(self, log_event):
         elif any(key in log_message for key in utils.LOG_PATTERN_IGNORE_LIST):
             return # Skip known noisy log patterns
 
+        # When a BulkExecutorError has been seen, suppress subsequent Glue exception analysis noise
```
hunterhacker, line 108: Could you please add a few more lines about what's the purpose behind all this, for us in 12 months when we've all forgotten? :)

tools/bulk_executor/server/src/python_modules/load_export/transform/.gitignore
```diff
@@ -0,0 +1,12 @@
+# Ignore all transforms by default (test transforms, personal transforms, etc.)
```
hunterhacker: Is this the right thing to do?

tools/bulk_executor/server/src/python_modules/load_export/transform/load_only_active.py
```diff
@@ -0,0 +1,57 @@
+from ..parsers.records import FullExportRecord, IncrementalExportRecord
```
hunterhacker, line 1: Should this be called "only-load-active.py" or something? When we had one example this was example.py but now we have several so seems like they each should be named what they do.

tools/bulk_executor/server/src/python_modules/load_export/transform/pii_mask_attribute.py
```diff
@@ -0,0 +1,116 @@
+from ..parsers.records import FullExportRecord, IncrementalExportRecord
```
hunterhacker, line 1: Seems appropriate to add a comment up top about what this shows.

"This example demonstrates how to modify attributes in place while using the table schema to ignore modifying PK/SK values. Specifically this masks the values within a set of named PII attributes."

tools/bulk_executor/server/src/python_modules/load_export/transform/load_only_active.py
hunterhacker: Also like the others a friendly comment:

"This demonstrates how to filter a load to only perform the load for items with certain characteristics. Here only items with an active status are loaded."

tools/bulk_executor/server/src/python_modules/load_export/transform/pii_remove_attribute.py
```diff
@@ -0,0 +1,87 @@
+from ..parsers.records import FullExportRecord, IncrementalExportRecord
```
hunterhacker, line 1: This example shows how to modify a load to remove unwanted attributes from items. Here the logic removes named attributes which are known to contain PII that we don't want loaded, but not removing any PK/SK attributes based on the stated schema.

tools/bulk_executor/server/src/python_modules/load_export/transform/pkmd5_add_attribute.py
```diff
@@ -0,0 +1,76 @@
+from ..parsers.records import FullExportRecord, IncrementalExportRecord
```
hunterhacker, line 1: This example shows how to add a new attribute to an item based on values of other attributes. It stores a calculated md5 hash as a new attribute.

tools/bulk_executor/server/src/python_modules/load_export/validators/manifest_validator.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        table_arn = manifest_summary.get('tableArn')
+        if not table_arn:
+            error_msg = "Missing tableArn in manifest-summary.json"
+            log.error(error_msg)
+            raise ValueError(error_msg)
+        
+        # Extract table name from ARN: arn:aws:dynamodb:region:account:table/table-name
+        try:
+            if not table_arn.startswith('arn:aws:dynamodb:'):
```
hunterhacker: We shouldn't hard-code aws here. GovCloud is 'aws-us-gov', for example.

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+        'key_schema_result': key_schema_result,
+        'step': step,
+    }
+
+
+def _estimate_cost(table_info, manifest_data, key_schema_result, step):
+    """Estimate and log DynamoDB write costs."""
+    step += 1
+    log.info(f"Step {step}: Estimating DynamoDB write costs...")
+    avg_item_size = key_schema_result.get('avg_item_size', 0)
```
hunterhacker, line 89: Default of 0? Will that ever happen? Free loads? :)

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+    log.info(f"Parser of type {type(parser).__name__} returned successfully...")
+
+    records_rdd = all_lines_rdd.map(parser.parse_to_record)
+    return records_rdd, export_load_type, parser, total_expected_items, step
+
+
+def _apply_transform_stage(spark_context, records_rdd, export_load_type, parser, transform_name, key_schema, error_accumulator):
+    """Apply transform then resolve records into items. Returns items_rdd, transform_active, transformed_out_accumulator, transformed_in_accumulator."""
+    transform_active = bool(transform_name)
+    transformed_out_accumulator = spark_context.accumulator(0) if transform_active else None
```
hunterhacker: Would be good to add a comment saying what these accumulators are for

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+                result = transform_fn(record)
+            except Exception as e:
+                error_accumulator.add([f"Transform function raised an exception: {e}"])
+                return []
+
+            if not isinstance(result, list):
+                result = [result]
+
+            if not result:
+                transformed_out_accumulator.add(1)
```
hunterhacker: If there's no result we add one to the out accumulator. The logic here is not intuitively obvious.

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+                return []
+
+            transformed_in_accumulator.add(len(result))
```
hunterhacker: Then the in accumulator adds the length of the result. I would've thought "in" was the before-transform and "out" was the after-transform. This reads like the reverse?

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+        return item
+
+    items_rdd = records_rdd.map(_resolve_and_validate).filter(lambda x: x is not None)
+    return items_rdd, transform_active, transformed_out_accumulator, transformed_in_accumulator
+
+
+def _write(spark_context, items_rdd, export_load_type, table_name, rate_limiter_shared_config, monitor_options, error_accumulator, debug_accumulator, step):
+    """Repartition and write items to DynamoDB."""
+    step += 1
+    num_partitions = min(items_rdd.getNumPartitions(), spark_context.defaultParallelism * 2)
```
hunterhacker, line 192: We had a lot of convos about this. Let's put your summary of the final thinking in a comment here of why we're doing this logic exactly.

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+def _report(manifest_data, total_expected_items, written_items_accumulator, transform_active, transform_name, transformed_out_accumulator, transformed_in_accumulator, start_time):
+    """Log final summary."""
+    total_item_count = manifest_data['total_item_count']
+    written_count = written_items_accumulator.value
+
+    log.info(f"Successfully wrote {written_count:,} items to DynamoDB")
+    if transform_active:
+        log.info(f"Transform '{transform_name}': {transformed_in_accumulator.value:,} items produced, {transformed_out_accumulator.value:,} items excluded out of {total_item_count:,} total")
```
hunterhacker: Oh your in is items that are going IN to the database and OUT is items kept OUT of the database. Since I first saw these around the transform logic I assumed it was in/out of the transform logic. Some comments when the variables are defined would help the reader. Naming as excluded/produced instead of out/in might be wise too.

tools/bulk_executor/server/src/python_modules/load_export/transform/default.py
```diff
@@ -0,0 +1,11 @@
+from ..parsers.records import FullExportRecord, IncrementalExportRecord
```
hunterhacker, line 1: Want to include an example that fans out an item to return a list of items? Or do we downplay that ability?

What's the most canonical example for why to do that? An item has an attribute that's a list called "regions" and you expand that out to have new items with "region" for each of the source regions? Maybe you have a better use case?

tools/bulk_executor/server/src/python_modules/load_export/README.md
```diff
@@ -0,0 +1,78 @@
+# Bulk Load-Export Capability
+
+This utility allows you to load a DynamoDB table exported to S3 by leveraging Glue. The code is modularized to enable unit testing of individual components. It also leverages the `RateLimiter` classes to ensure that any bulk action executed on a table only consumes the capacity configured.
```
hunterhacker: This is public-facing docs for the non-developer right?

Given the context, I'd not say "this utility" because this is a command inside Bulk.

I'd not say "by leveraging glue" because everything around here does that.

"This command loads a DynamoDB table export to an existing table."

tools/bulk_executor/server/src/python_modules/load_export/README.md
```diff
@@ -0,0 +1,78 @@
...
+def transform_incremental_record(record: IncrementalExportRecord) -> list[IncrementalExportRecord]:
+    # Add a 'migrated' flag to all items being inserted or updated
+    if record.new_image:
+        record.new_image["migrated"] = True
+    return [record]
+```
+
+**Templated Examples**
+
+The following ready-to-use transform modules are included in the `transform/` folder:
```
hunterhacker: I wouldn't say "ready to use" because they really aren't. They're demonstrations of technique. Like the PII ones require you edit them to specify attribute names.

tools/bulk_executor/server/src/python_modules/load_export/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- **`example.py`** - Filter items by attribute value, only loading items that match a condition (e.g. `status == "active"`).
+- **`pkmd5_add_attribute.py`** - Add an MD5 hash of the partition key as a new `pk_md5` attribute, useful for generating a deterministic hash-based attribute for any table.
+- **`pii_remove_attribute.py`** - Strip PII attributes (e.g. `Name`) from items during load, skipping any that are the partition key or sort key.
+- **`pii_mask_attribute.py`** - Mask PII attributes by keeping the first and last character of each word and replacing the middle with `*` (e.g. `"Alice Smith"` → `"A***e S***h"`), skipping any that are the partition key or sort key.
+
+## Role requirements
+The bulk load-export reads data from S3 and writes to an existing DynamoDB table, therefore it needs the following permissions:
```
hunterhacker: Where the "it" here is the Glue job role, not the caller role. You might need to change the bootstrapped role.

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/__init__.py
```diff
@@ -55,8 +57,9 @@ class RateLimiterWorker:
         shared_config (RateLimiterSharedConfig): The shared config between Aggregator and Worker.
         monitor_options: The expected monitor options (see @table_info#get_dynamodb_throughput_configs for more info)
     """
-    def __init__(self, shared_config, **monitor_options):
+    def __init__(self, shared_config, debug_accumulator=None, **monitor_options):
```
hunterhacker: Is the use of debug_accumulator no longer required here?

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorAggregator.py
```diff
@@ -2,15 +2,13 @@
...
-import sys
 import random
 from botocore.config import Config
 from botocore.exceptions import BotoCoreError, ClientError
 from concurrent.futures import ThreadPoolExecutor
 
 # Custom Library Imports
-sys.path.append('/server/src')
-from python_modules.shared.logger import log
+from ..logger import log
```
hunterhacker, line 11: Thank you for the path cleanup!

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
```diff
@@ -116,6 +116,9 @@ def _sync_loop(self):
                     # Smooth adjustment to avoid oscillation
                     #smoothing_factor = 0.2  # 0=no change, 1=jump immediately
                     smoothing_factor = 0.4  # 0=no change, 1=jump immediately
+                    old_write_rate = self.monitor.max_write_rate
+                    old_read_rate = self.monitor.max_read_rate
```
hunterhacker: Guessing these are leftovers from when you were debug_accumulator calling? Now they're not used.

tools/bulk_executor/tests/client/run_tests.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
hunterhacker, line 1: Should we have a comment that we expect this script to be run on a Mac? It's the first time we have an OS dependency.

tools/bulk_executor/tests/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Setup
+1. Activate the python virtual environment `source .venv/bin/activate`
+2. Install top-level python requirements `pip install -r requirements.txt`
+3. Move to the client test directory `cd tests/client`
+4. Running tests
+   1. Run all tests `./run_tests.sh -v --tb=short`
+   2. Run specific test for example `./run_tests.sh test_validate_s3_path.py -v`
+
+## Server testing
```
hunterhacker, line 14: Would be good to note it doesn't actually use any AWS resources for this.

tools/bulk_executor/README.md
```diff
@@ -504,6 +507,20 @@ If doing cross-account, you need a resource-based policy to enable access. The f
...
+* Does a full/incremental load of a DynamoDB table which was exported to S3 (i.e. currently only supports DDB-JSON format)
+* Requires `table` and `s3-path`
+* Validates manifest files to ensure integrity of the exported files and data
+* Supports loading a full export into an empty table or table with data (will overwrite existing data if keys match)
+* Supports loading an incremental export, will add new data and only update/delete data if keys match
+* Best used with:
+  * `--XMaxWriteRate N` where _N_ is the max aggregate write units per second to consume on the destination table
+  * `--transform` where you specify the python module used to transform items during load
+  * `--XTimeout` for large loads, set this appropriately to ensure Glue job doesn't time out, with a maximum of 10080 minutes (7 days)
+* Note: If there is an updated item in an incremental export and that item does not exist in the destination table, the tool will insert this item
```
hunterhacker: "the tool" -> "this command"

tools/bulk_executor/server/src/python_modules/load_export/validators/key_schema_validator.py
```diff
@@ -0,0 +1,116 @@
...
+
+            error = self._check_incremental_keys(data, expected_keys, i) if is_incremental else self._check_full_keys(data, expected_keys, i)
+            if error:
+                failed_rows.append(error)
+            else:
+                validated_count += 1
+                total_item_size += len(line.encode('utf-8'))
+
+        if failed_rows:
+            error_summary = f"Key validation failed for {len(failed_rows)} of {len(rows_to_check)} sampled row(s)"
```
hunterhacker, line 75: This error summary should include at least the first failed_rows error summary for the user to just see what went wrong.

tools/bulk_executor/server/src/python_modules/load_export/validators/key_schema_validator.py
```diff
@@ -0,0 +1,116 @@
...
+                else:
+                    actual_keys = first_row.get('Item', {})
+                actual_key_attrs = ', '.join(f"{k} ({list(v.keys())[0] if isinstance(v, dict) else '?'})" for k, v in list(actual_keys.items())[:10])
+                log.error(f"  Export data contains attributes: {actual_key_attrs}")
+            except Exception:
+                pass
+
+            for f in failed_rows:
+                log.error(f"  - Row {f['row']}: {f['error']}")
+            raise ValueError(f"{error_summary}. See logs for details.")
```
hunterhacker: Anytime we say "See logs" the user won't know what logs we're talking about.

tools/bulk_executor/server/src/python_modules/load_export/validators/key_schema_validator.py
```diff
@@ -0,0 +1,116 @@
...
+        rows_to_check = lines[:sample_size]
+        validated_count = 0
+        failed_rows = []
+        total_item_size = 0
+
+        for i, line in enumerate(rows_to_check):
+            try:
+                data = json.loads(line)
+            except json.JSONDecodeError:
+                failed_rows.append({'row': i, 'error': 'Malformed JSON'})
```
hunterhacker: Suggest including the line that caused the error 

```
'line': line
```

tools/bulk_executor/README.md
```diff
@@ -109,6 +109,9 @@ Here are some example use cases:
 # If going cross-account, you need a resource-based policy on the table to allow access.
 ./bulk copy --source source --target target
 ./bulk copy --source arn:aws:dynamodb:us-east-1:123456789012:table/source --target arn:aws:dynamodb:us-west-2:987654321098:table/target
+
+# Load a DynamoDB export into an existing DynamoDB table
+./bulk load-export --table target --s3-path "s3://..." [--transform example]
```
hunterhacker: Since people will prob try "s3://bucketname/prefix/" and not the whole path, we should make this example show something kinda realistic like:

```
 --s3-path "s3://bucketname/prefix/AWSDynamoDB/12345678901234/abcdef01/"
```

tools/bulk_executor/server/src/python_modules/load_export/transform/.gitignore
ruskindantra: IMO it is, this folder will be a very well curated list of transforms which are committed to the repository. So the "admin" has to make a conscience decision on which ones to commit and which ones not to, editing this file as needed when they are ready to commit that transformer for others to use.

tools/bulk_executor/server/src/python_modules/load_export/transform/default.py
ruskindantra: Can add lots of examples, later, as needed. The regions example is good, think denormalizing data into multiple rows.

tools/bulk_executor/server/src/python_modules/load_export/transform/load_only_active.py
ruskindantra: Agreed

tools/bulk_executor/server/src/python_modules/load_export/validators/key_schema_validator.py
ruskindantra: Just a little below this, we log the first export row and the failed rows.

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
ruskindantra: No, defensive coding, will never be 0. Upstream code will throw error before this is 0.

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
ruskindantra: Added a comment

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
ruskindantra: No, it's about which items are left (IN) and which items were removed (OUT). And because when items are left in, with a fan out operation, there could be a list

ruskindantra: I have added a comment further up to make it obvious

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
ruskindantra: Renamed

tools/bulk_executor/server/src/python_modules/load_export/transform/.gitignore
hunterhacker: Someone can commit to their own fork for all I care. They have to do a PR and get it approved for it to impact anyone else from the official repo. Someone who knows git will know how to "git add" or not, if they write their own file in there. Seems like there's a "principle of least surprise" that we're violating here.

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/__init__.py
ruskindantra: Correct, removed

tools/bulk_executor/server/src/python_modules/shared/rate_limiter/DistributedDynamoDBMonitorWorker.py
ruskindantra: Yes, removed

tools/bulk_executor/tests/client/run_tests.sh
ruskindantra: This will run on WSL too but I see your point. There is no PS (Powershell) equivalent which can be added later for posterity (if one exists)

ruskindantra: Something like 

```
 $Dir = Split-Path -Parent $MyInvocation.MyCommand.Path
 $ClientSrc = Join-Path $Dir "../../client/src"
 $env:PYTHONPATH = "$ClientSrc;$env:PYTHONPATH"
 python -m pytest $Dir @args
```

ruskindantra: Added a comment in the readme

tools/bulk_executor/README.md
ruskindantra: Okay will remove

tools/bulk_executor/server/src/python_modules/load_export/__init__.py
```diff
@@ -0,0 +1,364 @@
...
+    avg_item_size = key_schema_result.get('avg_item_size', 0)
+    estimated_size_bytes = avg_item_size * manifest_data['total_item_count']
+    get_and_print_table_write_cost(table_info, manifest_data['total_item_count'], estimated_size_bytes)
+    return step
+
+
+def _read_and_parse(spark_context, manifest_data, path_resolver, key_schema, step):
+    """Read export files into record RDD. Returns records_rdd, export_load_type, parser, total_expected_items, step."""
+    step += 1
+    log.info(f"Step {step}: Resolving export file paths...")
```
hunterhacker: The standard output should be limited to:

* The usual startup stuff (time, existence of ^C, etc)
* Quick metrics on what was read from S3 (like: number of files, items)
* Metrics on what happened during the transform (like: how many items now)
* The resulting cost estimate (based on the above)
* That writing has started (to give people a feeling to be patient)
* The final metrics that were written or errored out

People don't need to see these lines (I copied/pasted ones that don't seem generally useful):
```
Initializing...Bucket:aws-glue-bulk-dynamodb-us-east-1-654654401288-pdkgpacg, Prefix:aws-glue-bulk-dynamodb-us-east-1-654654401288-pdkgpacg
monitor_options {'aggregate_max_write_rate': 240000}
S3 Source Bucket: load-export-test-654654401288-us-east-1-an
S3 Source Bucket Prefix: entity-boat
S3 Source Bucket Export ID: 01777532144504-c0c0f215
Export Path: s3://load-export-test-654654401288-us-east-1-an/entity-boat
================================================================================
Step 2: Validating S3 export path exists...
S3 path validation successful: s3://load-export-test-654654401288-us-east-1-an/entity-boat
Step 3: Validating destination table exists...
Destination table validation completed successfully: {'pk': {'name': 'pk', 'type': 'S'}}
Step 4: Validating and parsing manifest files...
Starting manifest validation for export path: ExportPathResolver(bucket='load-export-test-654654401288-us-east-1-an', prefix='entity-boat', export_id='01777532144504-c0c0f215')
base_path s3://load-export-test-654654401288-us-east-1-an
data_base_path s3://load-export-test-654654401288-us-east-1-an/entity-boat
manifest_base_path s3://load-export-test-654654401288-us-east-1-an/entity-boat/AWSDynamoDB/01777532144504-c0c0f215
Manifest validator-1: Validating manifest-summary.json MD5 checksum...
manifest-summary read successfully
manifest-summary.md5 read successfully
manifest-summary.json MD5 checksum validated successfully
Manifest validator-2: Parsing manifest-summary.json...
Parsed manifest-summary.json: itemCount=479,023,691, outputFormat=DYNAMODB_JSON, exportType=FULL_EXPORT, exportTime=2026-04-30T06:55:44.504Z, exportFromTime=None, exportToTime=None
Manifest validator-3: Extracting table name from arn...
Extracted table name: entity-boat
Manifest validator-4: Validating output format...
Output format validated successfully
Manifest validator-5: Validating export views...
Manifest validator-6: Validating manifest-files.json MD5 checksum...entity-boat/AWSDynamoDB/01777532144504-c0c0f215/manifest-files.json
manifest-files.json MD5 checksum validated successfully
Manifest validator-7: Parsing manifest-files.json...
Parsed 1,024 data file entries from manifest-files.json
Manifest validator-8: Validating item count consistency...
Item count validated successfully: 479,023,691 items
Manifest validator-9: Manifest validation completed successfully: 479,023,691 total items across 1,024 data files
Step 4: Manifest validation completed successfully
Step 5: Validating data file checksums...
Validating MD5 checksums for 5 of 1,024 data files (sample mode)...
Data file MD5 validation completed: 5/5 files (sample mode)
Step 5: Data file checksum validation completed successfully
Step 6: Validating key schema against verified data files...
Key validation completed: 5/5 sampled rows verified (avg item size: 1,573 bytes)
Step 6: Key schema validation completed successfully
All validations passed successfully
Step 9: Resolving export file paths...
Total files to process: 1,024
Total expected items: 479,023,691
Step 10: Reading and parsing export files with Spark...
Parser of type FullExportParser returned successfully...
Loading transform module: remove_sk
Applying transform: remove_sk.transform_full_record
Step 11: Using 1,024 partitions (based on mins of #gz files 1,024 and defaultParallelism 876)
Step 11: Repartitioned to 1,024 partitions
Step 12: Writing items to DynamoDB in parallel...
Using batch writer for FULL_EXPORT load
```

And shouldn't the cost come after the transform, in case the transform suppresses a lot of items?

ruskindantra: 1. Cost calculation is done at the drivers. We don't know the outcome of the transforms until the DAG starts executing unfortunately. We can do an pre-estimate and a post-estimate. The pre-estimate gives the user a chance to exit early (cancel the job) if there are any surprises.
2. All the steps are using `log.info(...)`, could as well be easily changed to `log.debug(...)` although this is not something I agree with. Load export is an involved command and the information shown above is a clear indicator to the end user of:
    1. What is happening
    2. Where in the process they are and
    3. If things go wrong they know which step it happened in rather than the arduous process of passing in `--XDebug` and running the command again.

hunterhacker: If things go wrong, ideally the error message should be comprehensive enough to describe the error.

It would also be OK to accumulate the details and print them only if there's an exception. So say on validation error, print the steps that got to the error. But when things work, people don't need to know we "successfully read the manifest summary". That's just noise to the person.

ruskindantra: Addressed

Issue #142, opened 2026-04-30
Bulk's `diff` command has difficulty with some types


hunterhacker:
Testing diff on a table that has an item holding "one attribute with every data type", it generated this error:

```
  File \"/tmp/python_modules.zip/python_modules/diff.py\", line 196, in diff_segment
    if not item_matches(stream_a.head(), stream_b.head()):
  File \"/tmp/python_modules.zip/python_modules/diff.py\", line 120, in item_matches
    a, b = json.dumps(stream_a_item, sort_keys=True), json.dumps(stream_b_item, sort_keys=True)
  File \"/usr/local/lib/python3.10/json/__init__.py\", line 238, in dumps
    **kw).encode(obj)
  File \"/usr/local/lib/python3.10/json/encoder.py\", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File \"/usr/local/lib/python3.10/json/encoder.py\", line 257, in iterencode
    return _iterencode(o, 0)
  File \"/usr/local/lib/python3.10/json/encoder.py\", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

Cause is this code is too simplistic:

```
    a, b = json.dumps(stream_a_item, sort_keys=True), json.dumps(stream_b_item, sort_keys=True)
```

Probably needs to use the TypeSerializer:

```
from boto3.dynamodb.types import TypeSerializer
import json

serializer = TypeSerializer()

def normalize(item):
    return {k: serializer.serialize(v) for k, v in item.items()}

normalized = normalize(stream_b_item)
json.dumps(normalized, sort_keys=True)
```

There's a few other `json.dumps()` calls as well.

chrissamcat:
Kiro solution is to extend json.dumps to include a binary encoder. I am inclined to accept this over using the TypeSerializer as we can be confident how key sorting will work now and in the future, including in nested objects. Pushing fix imminently

hunterhacker:
Fixed by #143 (commit `3733dac`, "Fix: Handle DynamoDB binary attribute serialization in diff operations"), already on `main` — using the binary-encoder approach proposed above rather than `TypeSerializer`.

[...]

PR #143, opened 2026-05-01
Fix: Handle DynamoDB binary attribute serialization in diff operations

gllesch:
- Add BinaryAwareEncoder to convert bytes to base64 for JSON serialization
- Update item_matches() and log_diff() to use custom encoder
- Add comprehensive test suites for bug condition and preservation
- Fixes TypeError when comparing items with binary attributes

*Issue #, if available:*
https://github.com/awslabs/amazon-dynamodb-tools/issues/142

*Description of changes:*
Built and tested by Kiro 


hunterhacker:
Chris, do you mind moving the new test files (great to have!) under the ROOT/tests/* folder? It's a new folder that just got created in support of `load-export` via a PR I just merged in.

The reason I ask is everything under server/* gets zipped and put into the Glue context, and I want to minimize the surface area of things there to keep things maximally tightened down for security reasons.

gllesch:
Done!

ruskindantra:
Suggested changes

**Change 1**
The tests folder was designed to have folders per verb as each verb will have multiple unit tests. As each verb will be independently developed and tested, the folders can be run in their entirety.

So the diff test should be in `.../tests/[server|client]/diff/[testfile.py]`.

**Change 2**
Any specific instructions on how to run these unit tests? Please update the README.md in the tests folder as needed.

ruskindantra:
Awesome, thank you. Note that to run individual module unit tests only you will need to now do: `./run_tests.sh load_export --ignore=diff -v --tb=short` for example.

We can add this to the readme.md later, happy to merge

PR #159, opened 2026-05-22
Tests: add unit tests for the find verb

relentlesscol:
Adds unit tests for `python_modules/find.py` following the pattern in `tests/server/test_copy.py`.

## Coverage

**99% line + 98% branch on `python_modules/find.py`** (162 statements, 1 missed; 44 branches, 1 partial). 49 tests.

The single missed line is `find.py:95` — a defensive `raise ValueError(f"Invalid sort order: {order}")` that's structurally unreachable because the regex above already constrains `order` to `asc`/`desc`/empty. Reaching it would require monkeypatching the regex, which felt heavier than the test was worth. Worth a follow-up if a maintainer disagrees.

## Verification

```
.venv/bin/pytest tests/server/test_find.py --cov=python_modules.find --cov-branch --cov-report=term-missing
```

**Depends on #152.** This branch is cut off `bulk-executor-cleanup`. The diff vs `main` shows ~14 files because GitHub computes against `main`; the actual contribution is the single new test file. To see only the new test:

```
gh pr diff <PR> -- tools/bulk_executor/tests/server/
```

cc @hunterhacker for review

---


relentlesscol:
Rolled up into #152 (consolidated Makefile + tests PR). Closing in favor of that single PR.

Issue #145, opened 2026-05-04
Adopt the new Glue DynamoDB connector


hunterhacker:
Glue has a new DynamoDB connector as of Nov 2025: https://aws.amazon.com/about-aws/whats-new/2025/11/glue-dynamodb-connector/ 

This one goes straight to a DataFrame instead of using a DynamicFrame in between. This may resolve issue #81. It also has parameters to directly specify the read/write rates instead of requiring us to do the percentile math. This should also let us use the connector to write > 60k WRUs on on-demand tables.

More docs at:
https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-etl-connect-dynamodb-dataframe-support.html 

Make sure it works w the Glue versions we use, or that we can upgrade our Glue version.

hunterhacker:
Informally assigning to Colin (relentlesscol)

hunterhacker:
Done in #177

PR #162, opened 2026-05-27
bulk_executor: DataFrame-based DynamoDB connector + unit tests

### LINKED ISSUE #81: The toDF() call to convert a DynamicFrame to DataFrame can trigger an additional scan
Commands that use the DynamoDB Connector (`find`, `count`, `delete`, `sql`) internally create a `DynamicFrame` then call `toDF()` to convert it to a `DataFrame` in order to use the `DataFrame` features supporting `where`, `orderby`, and `limit`. This `toDF()` call will sometimes trigger a second table scan such as when the DynamicFrame couldn't be fully cached in memory. 

The cause is internal to the Connector and there's no simple workaround.

Could we just directly create a `DataFrame`? There's an open source project for this but we'd take a dependency on it: https://github.com/audienceproject/spark-dynamodb

Could we avoid the DataFrame? That's how `scancount` works. It uses a `FilterExpression` instead of a `where` clause. It doesn't need to support `orderby` or `limit` like a theoretical `scanfind` would have to.

Until there's a Connector fix, our cost estimates will take the pessimistic view that any invocation that calls of `toDF()` will do a second scan.

relentlesscol:
## Summary

Replaces the legacy DynamicFrame-based DynamoDB connector with the new DataFrame-based source AWS shipped November 2025
(<https://aws.amazon.com/about-aws/whats-new/2025/11/glue-dynamodb-connector/>).

**Why upgrade:** the new connector eliminates the issue #81 double-scan even when `cache()` spills (verified locally against a 506 GB table on both Glue 4.0 and Glue 5.0 with `--orderby` — legacy did 2.02× / 2.05× a full scan, new connector did 1.00×). For unordered `--limit` queries it pushes `LIMIT N` down into DynamoDB and reads ~N items instead of full-scanning.

## What's in this PR

Four logical commits:

1. **`Centralize Glue→DynamoDB connector behind a wrapper`** — `shared/glue_connector.py` provides three entry points (`read_dynamodb_dataframe`, `write_dynamodb_dataframe`, `count_dynamodb_table`) using the new DataFrame source. `find` / `sql` / `load` adopt them.
2. **`Bootstrap deploys Glue 5.0 + DYNAMODB connection`** — `GLUE_VERSION` 4.0 → 5.0; bootstrap upserts a `bulk-dynamodb-connection` Glue connection (idempotent `get_connection` → `EntityNotFoundException` → `create_connection`); job is created/updated with `Connections={'Connections': [name]}`. Teardown removes the connection after the job (Glue refuses to delete a connection still attached to a job). README IAM policy updated. `__version__` `0` → `1` so `assert_version_parity` trips for stale Glue 4.0 bootstraps.
3. **`Unit tests for the Glue connector wrapper`** — `tests/server/test_glue_connector.py` covers the wrapper's read/write/count branches and helpers; `tests/server/conftest.py` provides DataFrame-shaped wrapper mocks; `tests/client/test_bootstrap.py` and `tests/client/infrastructure/test_teardown.py` cover the new connection-management methods. Suite: 1,219 passing, 74 skipped (legacy-code-path tests pending wrapper-aware rewrite — see follow-ups below).
4. **`Tighten + complete bootstrap IAM policy`** — surfaced by an e2e security test (separate PR) that creates a temp IAM user, attaches the documented policy, and runs real bootstrap+teardown:
   - **Adds 3 missing actions:** `iam:PutRolePolicy` (bootstrap installs inline pricing/quotas policies), `iam:DeleteRolePolicy` + `iam:ListRolePolicies` (teardown removes them). Without these, a user following the README would succeed at bootstrap (just barely) and fail at teardown.
   - **Tightens scope:** the `wildcards` Sid used `Resource: "*"` for IAM role-admin actions — that lets the bootstrap principal manage *any* role in the account. Tightened to `arn:aws:iam::*:role/AWSGlueServiceRole*` (matching the existing `passrole` Sid's scope) and renamed the Sid to `glueRoleAdmin` so it's honest about what it grants.

## Verification

Verified locally end-to-end on `654654401288/us-east-1` against `another-bigger-boat` (1.7B items, 506 GB), `tiny-boat` (7.9M items), and `gdelt-col-data-mentions` (42M items, with `--orderby EventTimeDate`):

- ✅ Connector works on every verb tested
- ✅ Issue #81 double-scan eliminated on `find --orderby` (1.00× scan vs 2.02× legacy)
- ✅ Predicate-pushable workloads (`find --limit`, `delete --limit` without orderby) drop ~111-2,500× in RRU
- ⚠️ `count` without predicate is ~25% slower / more DPU than legacy's `DynamicFrame.count()` shortcut. Filed as Glue connector team feedback.

The e2e harness used to produce these numbers, plus the IAM security suite that surfaced the policy fixes in commit 4, are split out into a follow-up PR so the connector swap and its unit-test boundary land on their own.

## Jason's feedback (incorporated)

- Bumped client-server version identifier (`__version__` 0 → 1)
- Bootstrap creates the connection; teardown removes it
- README IAM policy updated with `glue:*Connection` permissions (commit 2) and the missing teardown actions + scope-tightening (commit 4)
- Removed the `--XConnectorVersion` switch — legacy is gone, dataframe is the only path

## Known follow-ups

- **E2e harness** lands as a follow-up PR once this merges (`bulk-executor-e2e-harness`).
- **`bulk bootstrap` is not idempotent on role-policy state.** It only attaches managed policies when *creating* the role. If a role exists but its policies have been detached (which the e2e security test triggers), re-bootstrap silently leaves it broken. Followup: re-attach managed policies on every bootstrap.
- **`--XJobName` override.** The security suite shares the `bulk_dynamodb` Glue job with the connector smoke suite, so running them against the same account causes interference. A `--XJobName` override on bootstrap/teardown/runner would let the security suite use a dedicated `bulk_dynamodb_security` job.
- **74 skipped unit tests** in `tests/server/test_{find,load,sql}.py` that asserted against the legacy DynamicFrame call chain. Followup: rewrite to assert against the wrapper boundary.
- File a Glue-connector-team ticket on the `count`-without-predicate regression.

Refs #145, fixes #81 (on the new connector path).

### DIFF
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -211,17 +211,20 @@ The bootstrap must be performed by a role with this policy at minimum:
     "Version": "2012-10-17",
     "Statement": [
         {
-            "Sid": "wildcards",
+            "Sid": "glueRoleAdmin",
             "Effect": "Allow",
             "Action": [
-                "iam:ListAttachedRolePolicies",
                 "iam:GetRole",
                 "iam:CreateRole",
                 "iam:DeleteRole",
                 "iam:AttachRolePolicy",
-                "iam:DetachRolePolicy"
+                "iam:DetachRolePolicy",
+                "iam:ListAttachedRolePolicies",
+                "iam:PutRolePolicy",
+                "iam:DeleteRolePolicy",
+                "iam:ListRolePolicies"
             ],
-            "Resource": "*"
+            "Resource": "arn:aws:iam::*:role/AWSGlueServiceRole*"
         },
         {
             "Sid": "passrole",
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -267,6 +267,19 @@ The bootstrap must be performed by a role with this policy at minimum:
             ]
         },
         {
+            "Sid": "glueConnection",
+            "Effect": "Allow",
+            "Action": [
+                "glue:CreateConnection",
+                "glue:GetConnection",
+                "glue:DeleteConnection"
+            ],
+            "Resource": [
+                "arn:aws:glue:*:*:catalog",
+                "arn:aws:glue:*:*:connection/bulk-dynamodb-connection"
+            ]
+        },
+        {
             "Sid": "logs",
             "Effect": "Allow",
             "Action": [
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -1,8 +1,8 @@
 # Bulk Executor for Amazon DynamoDB
 
-![tests](https://img.shields.io/badge/tests-1273%20passing-brightgreen)
-![line coverage](https://img.shields.io/badge/line%20coverage-99.7%25-brightgreen)
-![branch coverage](https://img.shields.io/badge/branch%20coverage-96.8%25-brightgreen)
+![tests](https://img.shields.io/badge/tests-1219%20passing-brightgreen)
+![line coverage](https://img.shields.io/badge/line%20coverage-95.4%25-brightgreen)
+![branch coverage](https://img.shields.io/badge/branch%20coverage-92.2%25-brightgreen)
 
 Bulk Executor for Amazon DynamoDB lets you efficiently run bulk commands against even large tables. It:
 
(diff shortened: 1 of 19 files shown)


relentlesscol:
Superseded by #177 (merged)

PR #172, opened 2026-06-10
bulk_executor: harden Glue 5.0 e2e coverage + fix connector write-path regressions

relentlesscol:
> **Draft for early feedback.** Branch is a few commits behind `awslabs:main` (#166/#169/#171 merged after it was cut) — there's a trivial additive conflict in `tests/client/test_runner.py` (this branch adds two jsonify test cases in the same region #171 touched). Easy rebase, will do before marking ready. Substantive work is all reviewable now.

## What this does

Verifies and hardens everything around the bulk_executor Glue 4.0→5.0 migration so the connector/security/command e2e suites and the PR #162 stack can be greenlit with confidence. While building out end-to-end command coverage, the verification surfaced **two real Glue 5.0 connector regressions** and a class of **false-green tests** — all fixed here, with regression guards.

## Real bugs found & fixed

**1. Connector write path broke on Glue 5.0 (two regressions, `load`-only)**
- `_looks_like_dataframe` misclassified Glue `DynamicFrame`s (they *do* expose `write`+`schema`), so the wrapper skipped `toDF()` and `df.write.format()` failed with `'function' object has no attribute 'format'`. Now detects a DynamicFrame **positively** via `hasattr(toDF)`.
- Spark's default `ErrorIfExists` save mode is rejected by the Glue 5.0 DynamoDB connector. Added `.mode("append")` (DynamoDB writes are upserts). 
- Only `load` hit these — `fill`/`copy`/`update` write via boto3, not the Spark connector. Added regression guards (realistic DynamicFrame mock, real-DataFrame no-conversion test, `writer.mode('append')` assertion).

**2. e2e smokes were false-green** — `./bulk` exits 0 even when its Glue job *fails*. Smokes that asserted only on the CLI exit code passed against crashed jobs. New `assert_glue_succeeded()` checks the authoritative `JobRunState == SUCCEEDED` via `glue.get_job_run`, plus `assert_table_has_items()` to prove writes landed.

**3. Perf math** — DPU-seconds was `ExecutionTime × MaxCapacity` (worker ceiling), overstating cost ~100× under autoscaling. Now uses Glue's real `DPUSeconds` field.

**4. Security suite clobbered the shared Glue job** — the real-bootstrap tests tore down / flipped the shared `bulk_dynamodb` job to read-only, breaking subsequent write smokes. Added a net-zero snapshot/restore guard. Also fixed a false-negative where removing `glue:UpdateJob` didn't fail bootstrap (it's only exercised on re-bootstrap, not the fresh-account `create_job` path the test takes).

## Test coverage

- **New bulk-command e2e smoke suite** (`fill`/`update`/`delete`/`copy`/`diff`) — exercises the verb-specific Spark orchestration the connector smokes don't, the part most at risk from the 5.0 jump. Each creates a transient table, asserts the Glue job SUCCEEDED, and verifies real effects.
- **Rewrote 74 skipped unit tests** that were disabled during the 4.0→5.0 migration (`@pytest.mark.skip` "rewrite later"). Skipped tests look green but cover nothing and erase the contracts they documented. Each is now rewritten against the current wrapper boundary, or deleted with an explicit note pointing to where coverage moved (`test_glue_connector.py`). **Result: unit suite 1292 passed, 0 skipped, 0 skip markers.**
- `verb`→`command` terminology rename across the e2e harness (matches the rest of the codebase).

## Verification (against the bulk shared account, Glue 5.0)

| Suite | Result |
|-------|--------|
| Unit | **1292 passed, 0 skipped** |
| Command e2e (`make test-e2e-commands`) | **5/5** |
| Connector e2e (`make test-e2e-connector`) | **4/4** |
| Security e2e (`make test-e2e-security`) | **9/9** |

## Review notes for @ (reviewer)

- Commits are organized by concern (connector fix / unit-test rewrite / truthful assertions / command suite + rename / security / jsonify) — reviewable commit-by-commit.
- The `load`-path fixes (commit 1) are the highest-risk change — that's the one Glue 5.0 write path; please scrutinize `_is_dynamic_frame` and the `append` mode.
- The 74-test rewrite (commit 2): the deletions all carry in-file rationale + pointers; happy to walk through any that look like lost coverage.


relentlesscol:
Superseded by #177 + #186

Issue #92, opened 2025-10-14
Have `scancount` able to output item counts per segment to measure possible task execution skew


hunterhacker:
It would be useful sometimes if a `scancount` run could print the number of (matching) items per segment instead of overall. This would show if the data was significantly skewed. Imagine you have 5 segments and for each you have an item count:

0: 1000
1: 900
2: 1000
3: 123456789
4: 1000

This idea comes from someone who on a 7m item table had 5m items with the same "null" pk value. This caused the `update` call to be slow because it ended up essentially doing single-threaded updates on that one segmented scan segment.

Perhaps:

```
./bulk scancount --table t --persegment --segments 200
```

So says to do 200 segments and get a count per segment.

Perhaps there's better ways to measure skew? 

Really what we care about here is task duration. You'd check this if things were behaving weirdly. If we divide the table into 200 segments and give 1 segment to a task, will they each do about 0.5% of the work or will one get 25% of the data? 

OK, technically item count doesn't measure skew as well as summed item size, but item count is probably a decent stand-in.

relentlesscol:
Taking another look

hunterhacker:
Implemented — `scancount --per-segment` prints per-segment item counts to reveal task/partition skew. Client flag: `client/src/python_modules/scancount.py` (`--per-segment`); server: `server/src/python_modules/scancount/__init__.py` (`per_segment` → per-segment count reporting). Commit `04ab7d3` ("scancount: single-pass --per-segment + sampled estimate").

PR #190, opened 2026-06-25
[bulk_executor] feat: scancount --per-segment flag to reveal data skew

relentlesscol:
## Summary

Adds a `--per-segment` mode to the `scancount` command that prints item counts per DynamoDB scan segment instead of only the total. Makes it easy to spot partition skew — e.g. one segment holding 5M items while others have 1K.

Also adds a `--segments` flag (default: 200) so users can control the number of parallel scan segments. Pairs naturally with `--per-segment` for skew analysis at different granularities — e.g. `--segments 10` for a quick overview on a small table, or the full 200 for high-resolution skew detection on large ones.

When `--per-segment` is set, after the normal total-count run completes, a second pass uses `rdd.map().collect()` to gather `(segment, count)` tuples. Results are printed sorted by count descending with percentage-of-total and basic statistics (mean, skew ratio). A warning is emitted when the max/mean ratio exceeds 5x, indicating significant hot-partition risk.

Addresses #92

## Changes

- `--per-segment` flag on `scancount` to print per-segment item counts
- `--segments N` flag to control parallel scan segment count (default 200, backwards compatible)
- Skew statistics: mean, max/mean ratio, warning when ratio > 5x

## Testing

1263 passed, 49 skipped (94.7% line coverage, 92.2% branch coverage)

### DIFF
--- a/.github/scripts/e2e-bootstrap.sh
+++ b/.github/scripts/e2e-bootstrap.sh
@@ -1,0 +1,238 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Bootstrap GitHub Actions e2e infrastructure in a given AWS account.
+# Creates: OIDC identity provider, IAM role, inline policy, GitHub secrets.
+#
+# Prerequisites:
+#   - AWS CLI authenticated to the target account (Admin role)
+#   - gh CLI authenticated with repo admin access
+#   - jq installed
+#
+# Usage:
+#   .github/scripts/e2e-bootstrap.sh \
+#     --account-id 123456789012 \
+#     --region us-east-1 \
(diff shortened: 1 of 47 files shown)


hunterhacker:
Seems like if we do this maybe we should let the user control the number of segments we're using. Right now it's hard-coded as 200 because that's good enough for the goal of parallel scan-based counting (200 clients each doing rate limited 1,500 RCUs = 300,000 RCUs which is above what most people would require) but with this change maybe people would want 10 segments or 100 segments?

Issue #92 mentioned a --segments parameter.

hunterhacker:
Hey AI, shouldn't this include an update to the README.md?

relentlesscol:

Issue #83, opened 2025-10-14
Quicker exit on non-recoverable systemic errors


hunterhacker:
Right now, if one task gets a fatal error, we're adding it to an `error_accumulator` and then in the driver pulling that error out for display once all executors finish. Really though, once a non-recoverable system-wide fatal error has been hit, it might be nice to let other tasks to know to exit early on startup or ideally in-place for long-running work. There's not a nice way in Spark to do this in-band.

One option is in our S3 bucket (already used for various things including rate limiting) we could have a kill-switch object which tasks can look for to know "oh we're done here" and early exit, thus not wasting time running once a fatal error has been found.

For example, this code inside `fill` should do something after adding the error to the accumulator to not only end this task but cause all other tasks to stop.

    except botocore.exceptions.ClientError as e:
        if get_error_code(e) == DYNAMO_DB_THROTTLE_EXCEPTION:
            print('Persistent throttling on batch_writer exit, give up on last few item inserts...')
        elif get_error_code(e) == DYNAMO_DB_VALIDATION_EXCEPTION:
            error_accumulator.add([f"Schema validation error: Generated items don't match table schema"])
            # Must all tasks hit the same error? It doesn't take long, but it seems messy.
        else:
            error_accumulator.add([f"Error during writing: {get_error_message(e)}"])

Alternatively, we could output to CW a sentinel string saying effectively "--- Well I'm dead with an exception ---" (exact text TBD) and use our existing code that watches the logs looking for fatal issues to trigger an early job stoppage.

PR #197, opened 2026-06-29
[bulk_executor] feat: early-exit on non-recoverable systemic errors

relentlesscol:
Addresses #83. Adds poison-pill detection for non-recoverable errors (IAM denials, missing table, etc.) so the job exits quickly instead of retrying until timeout.

~400 lines (error-classification logic + tests).


relentlesscol:

PR #199, opened 2026-06-29
[bulk_executor] fix: require PITR enabled for load command

### LINKED ISSUE #179: "Load" should require pitr enabled
Noticed that the `load` command doesn't require PITR be enabled. It should be passing `pitr_enabled` to the validate call to indicate it's a mutator.

```
    utils.validate_tables(env_configs, parser, result['table'], pitr_enabled=True)
```

relentlesscol:
Fixes #179.


relentlesscol:

PR #201, opened 2026-06-29
[bulk_executor] feat: report write rate at load start

### LINKED ISSUE #182: The "load" command makes no statements about write rates

relentlesscol:
Fixes #182.


relentlesscol:

PR #205, opened 2026-06-29
[bulk_executor] feat: add --XExistingBucket parameter to bootstrap

### LINKED ISSUE #130: Allow bootstrap to specify an existing S3 bucket
Some customers may not be able to give the bootstrap context enough permissions to create an S3 bucket. They might like to pass in an existing S3 bucket. 

Now, for IAM to allow the Glue job role to work with the bucket, we're necessitating the bucket name to follow the Glue naming prefix convention. Easiest approach then is if the passed-in bucket can use the right name. If not, the solution would be to give the Glue job a role that allows access to the bucket. We'd solve this in docs, presumably.

relentlesscol:
Fixes #130.


relentlesscol:

PR #206, opened 2026-06-29
[bulk_executor] feat: warn if custom --XRole lacks minimum permissions

### LINKED ISSUE #82: Check any passed-in custom role has required minimum permissions
When someone specifies a custom role to use during bootstrap, we already sanity check that it exists but would be good to also check it has the required minimum permissions (as specified in the documentation). 

If by chance it doesn't, the bulk execution will fail in possibly subtle ways like not being able to write to CloudWatch and so not generating output.

relentlesscol:
Fixes #82.


relentlesscol:

PR #207, opened 2026-06-29
[bulk_executor] feat: force built-in role refresh on version mismatch

### LINKED ISSUE #84: Force built-in role refresh on version mismatch
Every time we `bootstrap` we do NOT overwrite the generated roles. But yet sometimes we add permissions to them. That can leave the user with a permission missing error where the role they first bootstrapped isn't as capable as it needs to be for the new version.

It would be inefficient (for us as devs anyway) to always regen the role, but we could regen the role if the `__version__.py` protocol version number was incremented.

Workaround today is a teardown/bootstrap combo.

Definitely implement this feature if ever expanding the permissions built into the generated roles.

relentlesscol:
Fixes #84.


relentlesscol:

PR #208, opened 2026-06-29
[bulk_executor] feat: add --XIdleTimeout for Glue cost optimization

### LINKED ISSUE #88: Consider a shorter Glue `IdleTimeout` for cost optimization
By default Glue workers have a 10-minute idle timeout. This is configurable through the `IdleTimeout` parameter. If we set it to a smaller value it could perhaps save costs. Theoretically it would be beneficial when one task is still running but all the rest are idle. Theoretically it would be worse to have a shorter `IdleTimeout` if there were a second burst of activity after the idle time, but that's not something I'd expect common for bulk commands.

It would go in the same place as the other Glue params:

```
{
    ...
    "WorkerType": "G.1X",
    "NumberOfWorkers": 2,
    "GlueVersion": "3.0",
    "Name": "your-job-name",
    "Role": "your-role-arn",
    "Command": {...},
    "ExecutionClass": "STANDARD",
    "IdleTimeout": 2
}
```

Suggest creating a test case where one executor runs longer than others, confirming DPU cost, doing this change, confirming lower DPU cost.

relentlesscol:
Fixes #88.

### DIFF
--- a/tools/bulk_executor/client/src/infrastructure/constants.py
+++ b/tools/bulk_executor/client/src/infrastructure/constants.py
@@ -38,6 +38,7 @@ class GlueJobDefaults(Enum):
     MaxConcurrentRuns=20
     Retries=0
     Timeout=60
+    IdleTimeout=5
     NumberOfWorkers=220
     WorkerType='G.1X'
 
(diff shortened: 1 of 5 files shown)


relentlesscol:

PR #209, opened 2026-06-29
[bulk_executor] feat: add rate validation warnings for XMaxReadRate/XMaxWriteRate

### LINKED ISSUE #89: Expand warnings if requested read/write rate is too high or too low
Today we put in a rule that if the determined read/write rate is < 100 we'll warn the user. We can do more.

First, there's the case where they're going to run above 100 but it's still too low. Look at the metadata on the table, decide how long the read/write rate will take. If it's too long (over an hour which is our default job duration), emit a warning to that effect. So if the job will run at 1k but the table has a billion items, your Glue job will timeout before it's all read. There's subtlety here in if the 1k is set by the user or determined by the table's settings, etc.

Second, there's the case where they specifically ask for more than the table can do. For example, if someone specifies a `XMaxReadRate` or `XMaxWriteRate` and the table can't do it, we should probably emit a warning. For example:

* If the table doesn't have autoscaling and is provisioned below their request.
* If the table does have autoscaling but the max is below their request.
* If the table is on-demand but the max setting on the table is below their request.
* If the table is on-demand and there's no max setting on the table but the account quota is below their request.

We might give a different warning if:
* Table is provisioned below the request but auto-scaling will be able to adjust upward to match the request.

relentlesscol:
Fixes #89.

### DIFF
--- a/tools/bulk_executor/server/src/python_modules/shared/table_info.py
+++ b/tools/bulk_executor/server/src/python_modules/shared/table_info.py
@@ -11,6 +11,45 @@ from .pricing import PricingUtility
 MIN_RECOMMENDED_READ_RATE = 100
 MIN_RECOMMENDED_WRITE_RATE = 100
 
+# Maximum multiplier over table capacity before we warn the user.
+# For provisioned tables, exceeding 100% will cause throttling.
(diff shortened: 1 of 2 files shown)


relentlesscol:

PR #210, opened 2026-06-29
[bulk_executor] perf: install faker only for fill verb to speed worker startup

### LINKED ISSUE #91: Enable `faker` only for `fill` to speed up startup
Looking at the worker logs, we see 8-15 seconds to install faker even for jobs that aren't `fill` jobs.

If we only included `faker` for `fill` jobs, we'd probably have faster worker spin-up times.

Marking as a feature because things work fine now, but this is a speed optimization. 

Probably have `--additional-python-modules` pointing at `faker` added only to `fill` runs.

relentlesscol:
Fixes #91.


relentlesscol:

PR #212, opened 2026-06-29
[bulk_executor] Warn if XTimeout seems insufficient based on item count and rate

### LINKED ISSUE #144: Dynamically report if `XTimeout` seems insufficient
We have a 60 minute timeout. Sometimes it will be obvious the job won't complete in that time, like if there's 2 billion items and a 10,000 XMaxReadRate, in which case we could be polite and warn the user early as part of the cost estimate.

relentlesscol:
Fixes #144.

Developed and tested on fork branch `polecat/bu-i144`. Merges cleanly onto main.


relentlesscol:

PR #214, opened 2026-06-29
[bulk_executor] fix(tests): resolve rate_limiter import shadow breaking test collection

relentlesscol:
## Summary
- Removes `__init__.py` from `tests/server/rate_limiter/` so it no longer shadows the source module `python_modules.shared.rate_limiter`
- Rewrites the conftest to only register leaf submodule entries in `sys.modules` without replacing the parent rate_limiter Mock, preventing pollution of other server tests

## Problem
Running the full test suite (`make test`) caused 8 collection errors because `tests/server/rate_limiter/__init__.py` turned the test directory into a package. Its conftest replaced `sys.modules['python_modules.shared.rate_limiter']` with a partial real module that lacked the wrapper classes (`RateLimiterAggregator`, `RateLimiterWorker`), causing `ImportError` in all other server tests that import from that module.

## Test plan
- [x] `make test` passes: 1404 passed, 0 collection errors
- [x] All 74 rate_limiter unit tests still pass
- [x] No regression in other test suites

### DIFF
--- a/.github/scripts/e2e-bootstrap.sh
+++ b/.github/scripts/e2e-bootstrap.sh
@@ -1,0 +1,238 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Bootstrap GitHub Actions e2e infrastructure in a given AWS account.
+# Creates: OIDC identity provider, IAM role, inline policy, GitHub secrets.
+#
+# Prerequisites:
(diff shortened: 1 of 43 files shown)


relentlesscol:
Closing: opened from stacked branch instead of fork/main. Replaced by a single clean PR from fork's integration branch.

Issue #184, opened 2026-06-25
Should be able to round trip a "find" writing to S3 and a "load" loading from S3


hunterhacker:
I did a test on a table with 3 items trying to use every data type. Here's the `aws scan` output of the test data:

```
{
    "Items": [
        {
            "imamap": {
                "M": {
                    "listinside": {
                        "L": [
                            {
                                "N": "10"
                            },
                            {
                                "N": "11"
                            }
                        ]
                    }
                }
            },
            "imabool": {
                "BOOL": false
            },
            "imabs": {
                "BS": [
                    ""
                ]
            },
            "s": {
                "S": "y"
            },
            "imalist": {
                "L": [
                    {
                        "N": "0"
                    },
                    {
                        "N": "0"
                    },
                    {
                        "N": "0"
                    }
                ]
            },
            "imass": {
                "SS": [
                    "x"
                ]
            },
            "imanull": {
                "NULL": true
            },
            "imanum": {
                "N": "0"
            },
            "imans": {
                "NS": [
                    "0"
                ]
            },
            "p": {
                "S": "x"
            }
        },
        {
            "s": {
                "S": "hello"
            },
            "p": {
                "S": "one\ntwo and \\n three"
            }
        },
        {
            "b": {
                "BOOL": true
            },
            "bin": {
                "B": "fecabeba"
            },
            "num": {
                "N": "5"
            },
            "ns1": {
                "NS": [
                    "2",
                    "1"
                ]
            },
            "mylist": {
                "L": [
                    {
                        "S": "xx"
                    },
                    {
                        "N": "0"
                    },
                    {
                        "BOOL": false
                    },
                    {
                        "B": "feca"
                    },
                    {
                        "SS": [
                            "a",
                            "b"
                        ]
                    },
                    {
                        "L": [
                            {
                                "S": "hi from deep inside"
                            }
                        ]
                    }
                ]
            },
            "p": {
                "S": "a"
            },
            "bins1": {
                "BS": [
                    "beba",
                    "feca"
                ]
            },
            "payload": {
                "S": "alpha is here"
            },
            "emptybins1": {
                "BS": [
                    ""
                ]
            },
            "ss1": {
                "SS": [
                    "abc",
                    "def"
                ]
            },
            "s": {
                "S": "b"
            },
            "nope": {
                "NULL": true
            },
            "map1": {
                "M": {
                    "inner": {
                        "S": "innerval"
                    },
                    "moremaps": {
                        "M": {
[...]

Issue #94, opened 2025-10-15
[verb] `scanfind` - does parallel scans with a push-down predicate instead of using the connector


hunterhacker:
Today's `find` with a where clause does not do push-down predicates, meaning it pulls the whole table into memory, converts from a dynamic frame to a data frame (which can trigger a second scan sometimes), then applies the spark where filtering.

A more memory-efficient approach would be a verb that uses direct boto3 access to do scan calls with a FilterExpression. We already do something like this with `scancount`. So maybe the new verb is `scanfind`.

We could support a `limit` parameter easily enough. Doing `orderby` would be a trick.

Then this could be the basis for a `scandelete` where the delete has a ConditionExpression of the filtering criteria, to ensure it was still true at the time of delete. Today's Spark-driven delete works by finding matching items, then deleting them, and if an item changes to no longer match at the last moment it's still going to be deleted. If we filter using a ConditionExpression then it can be reused on the delete item.

I don't love `scancount`, `scanfind`, and `scandelete` as names. Still, we've established the precedent with `scancount`.

hunterhacker:
As #221 wanted, a big win of this over `find` is it would have perfect data type knowledge (the connector used by `find` is data type lossy) and so we could `--output ddb-json` in a way that could be reimported back in. Basically giving us a very nice filtered export.

Also `--output parquet` would be nice, and any other outputs we'd want. 

Curious if we'd wand to build up an RDD in each worker to manage the S3 output, esp for formats like Parquet.

PR #221, opened 2026-07-01
[bulk][split-217 #184] find: preserve DynamoDB types in S3 output (bu-nd9)

relentlesscol:
## Summary

Issue #184. Preserve DynamoDB types in find-to-S3 for round-trip with load. Files: server/src/python_modules/find/. Must pass make test.

## Implementation notes

Implemented: find verb now writes DynamoDB JSON (type-annotated) to S3, preserving SS/NS/BS/B/NULL/L/M for round-trip with load. New shared/ddb_json_writer.py module uses connector schema metadata. All 1356 tests pass.

## Refinery handoff

- Issue: `bu-nd9` (task, P2)
- Source branch: `polecat/bu-nd9`
- Target: `main`

### DIFF
--- a/tools/bulk_executor/client/src/python_modules/find.py
+++ b/tools/bulk_executor/client/src/python_modules/find.py
@@ -11,12 +11,17 @@ help_text=f"""
         Optional --where parameter to specify a match criteria, using Spark SQL syntax
         Optional --orderby parameter to specify a sort attribute, with optional asc/desc suffix
         Optional --limit parameter to limit the number of items processed
+        Optional --format parameter to specify the S3 output format:
+            json      - Plain JSON (default). Human-readable but loses DynamoDB type info.
+            ddb-json  - DynamoDB JSON with type descriptors. Round-trips with 'bulk load'
+                        and the native DynamoDB Import from S3 feature.
+            parquet   - Apache Parquet columnar format. Compact and fast for analytics.
         Saves full output to S3 and prints the top few items to console
 
     Examples:
         bulk find --table products
         bulk find --table users --where "age > 21"
-        bulk find --table orders --where "status = 'pending'"
+        bulk find --table orders --where "status = 'pending'" --format ddb-json
     """
 
 def validate_orderby(parser, values):
@@ -40,6 +45,7 @@ def run(env_configs, verb="find", help_text=help_text):
     parser.add_argument('--where', type=str, default=argparse.SUPPRESS, help='Where clause')
     parser.add_argument('--orderby', nargs='+', type=str, default=argparse.SUPPRESS, metavar=('COLUMN', 'DIRECTION'), help='Order by clause (e.g., "column" or "column asc/desc")')
     parser.add_argument('--limit', type=int, default=argparse.SUPPRESS, help='Limit number')
+    parser.add_argument('--format', type=str, default='json', choices=['json', 'ddb-json', 'parquet'], help='S3 output format (default: json)')
     args = parser.parse_args()
 
     if hasattr(args, 'orderby'):
--- a/tools/bulk_executor/server/src/python_modules/find.py
+++ b/tools/bulk_executor/server/src/python_modules/find.py
@@ -7,7 +7,6 @@ import warnings
 import boto3
 from awsglue.transforms import *
 from botocore.config import Config
-from pyspark.sql import SparkSession
 from pyspark.sql.functions import asc, desc
 
 # Custom Library Imports
@@ -24,6 +23,7 @@ from python_modules.shared.table_info import (
     get_and_print_dynamodb_table_info, get_and_print_table_scan_cost,
     get_dynamodb_throughput_configs)
 from python_modules.shared.glue_connector import read_dynamodb_dataframe
+from python_modules.shared.ddb_json_writer import write_ddb_json_to_s3, row_to_ddb_json
 
 
 def print_dynamodb_table_info(table_name, is_delete, **kwargs):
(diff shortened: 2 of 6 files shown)


hunterhacker:
This change adjusts the S3 output but doesn't make it configurable. I think we need it configurable. As the original ticket said:

Maybe `--format` with options of `ddb-json` or `parquet` or `json` which is today's default? The reason for `ddb-json` is it works with the native import from S3 feature as it exists today.

Interesting topic if `json` is the right default. It's what we have now, tho.

relentlesscol:
Addressed: added `--format` flag (json, ddb-json, parquet) to make S3 output configurable per your feedback. Commit: https://github.com/relentlesscol/amazon-dynamodb-tools/commit/55176ef

hunterhacker:
## e2e test: `--format ddb-json` does not preserve set/number types (real Glue run)

I ran this branch end-to-end against a real DynamoDB table with one item containing every type (S, N, B, BOOL, NULL, SS, NS, BS, L, M), via a live Glue job, and inspected the actual S3 output. The unit tests pass, but they feed hand-built `dynamodb.type` metadata that **the real connector doesn't provide the way the code assumes**, so several headline cases are wrong in practice.

**Put in → came out (from the written S3 file):**

| Attribute | Input | Output | |
|---|---|---|---|
| `strSet` | `{"SS":["alpha","beta","gamma"]}` | `{"L":[{"S":"alpha"},...]}` | ❌ SS → L |
| `numSet` | `{"NS":["1","2.5","-7"]}` | `{"L":[{"S":"-7"},{"S":"2.5"},{"S":"1"}]}` | ❌ NS → L, **N → S** |
| `binSet` | `{"BS":["AAE=","AgM="]}` | `{"L":[{"B":"AAE="},...]}` | ❌ BS → L |
| `list` | `[{"S":"x"},{"N":"9"},{"BOOL":true},{"NULL":true}]` | `[{"S":"x"},{"S":"9"},{"S":"true"},{"NULL":true}]` | ❌ **N → S, BOOL → S** |
| `nul` (top-level) | `{"NULL":true}` | *(attribute missing entirely)* | ❌ dropped |
| `map.tags` | `{"SS":["a","b"]}` | `{"L":[{"S":"a"},{"S":"b"}]}` | ❌ SS → L |
| S / N / B / BOOL / empty-string / map scalars | — | preserved | ✅ |

**Root causes:**
1. **Sets collapse to lists.** The connector surfaces DynamoDB S/N/B *sets* as Spark `ArrayType` and the `dynamodb.type` metadata the writer keys on (`"SS"/"NS"/"BS"`) isn't present as expected, so `_infer_ddb_type` falls back to `L`.
2. **Scalars inside collections stringify.** Elements inside sets/lists lose their type — numbers and bools serialize as `{"S": ...}`.
3. **Top-level `NULL` is dropped** — `row_to_ddb_json` does `if value is None: continue`, so a `NULL` attribute vanishes.

**Impact:** the stated purpose is round-tripping with `bulk load` / DynamoDB Import-from-S3. As-is, a number set round-trips as a list of strings and a `NULL` disappears — a silent data-fidelity change, not a faithful round-trip.

**Suggestion:** the fix needs to be driven by what the connector *actually* emits (metadata shape + how it represents sets), which only shows up in a real run — the current unit tests encode an idealized metadata contract that doesn't hold. Happy to help pair on a metadata-accurate version. (Separately: this branch predates the `pricing.py` `KeyError: 'group'` fix (#268), so cost estimation crashed until I patched it locally to run this test — worth rebasing on latest main.)

Test artifacts were `pr221-*` tables in the bulk-testing account; cleaned up.

relentlesscol:
**Root cause confirmed.** Dug into this and your live run holds up. The problem sits upstream of the writer: the DynamoDB DataFrame connector (`spark.read.format("dynamodb")`) that `find` uses drops set-ness and number precision at read time, and no read option brings them back.

What the connector's inferred schema gives us on a real run:
- Every field is a plain `StructField(name, dataType, nullable=true)` with no type metadata. The `dynamodb.type` key the writer keys on isn't present for this source.
- `SS`, `NS`, `BS` all infer to the same `ArrayType` as a homogeneous `List` — identical shape, no marker. Nothing in the DataFrame distinguishes a string set from a list of strings.
- `N` comes back as `Long`/`Double`, never `Decimal`, so precision can degrade.
- `simplifyDDBJson` only applies to the S3-export reader, not the live scan, so it doesn't help here.

Since `find` needs the live connector for WHERE/ORDERBY, SS→L can't be recovered on that path — the writer never sees the set tag.

Two things here are genuinely fixable and separate from the connector limitation:
1. Top-level `NULL` is dropped — `row_to_ddb_json` does `if value is None: continue`. Should emit `{"NULL": true}`.
2. Scalars inside lists stringify (`N`→`S`, `BOOL`→`S`). The connector gives us each element's Spark type, so those can serialize correctly.

Also confirmed your read on the unit tests: they feed hand-built `dynamodb.type` metadata the connector never emits, so they pass while the real output is wrong.

Parking this for now — happy to come back to it if it's still open. The realistic paths are shipping `ddb-json` for the types that survive (S/N/B/BOOL/M/L) with sets documented as degrading to `L`, or reading a PITR export as a boxed DataFrame to keep every tag at the cost of an export step.

relentlesscol:
Parking this PR in favor of #94 (`scanfind`).

Your e2e run is right, and the limitation sits upstream of the writer: `find` reads through the Spark DynamoDB connector, and once the data is a DataFrame the type information needed for a faithful round-trip is already gone.

- **Sets are unrecoverable on this path.** Spark has no set type, so `SS`/`NS`/`BS` infer to the same `ArrayType` as a list, and the live connector attaches no `dynamodb.type` metadata to distinguish them. No read option brings set-ness back.
- **Numbers** come back as `Long`/`Double`, never `Decimal`, so precision can degrade.
- The **top-level `NULL` drop** and the **scalar-in-list stringifying** were genuine writer bugs on top of that, but fixing them still leaves sets wrong.

Since the goal (#184) is a faithful round-trip into `load` / Import-from-S3, the connector path can't get there. `scanfind` (#94) can: a native segmented `scan` with a `FilterExpression` push-down returns wire-format DynamoDB JSON with every type intact, and it skips pulling the whole table through the dynamic-frame to data-frame conversion (and the occasional second scan) that `find --where` pays today. It parallelizes the same way `scancount` does.

Closing this in favor of #94. #184 stays open as the goal. The `--format` CLI wiring here is still reusable if we ever want `find --format json|parquet` for the non-round-trip analytics case, which isn't blocked by the connector limitation.

PR #229, opened 2026-07-02
[bulk][split-218 #82] Warn if custom role lacks minimum permissions (bu-iot)

relentlesscol:
## Summary

Issue #82. When user passes --XRole, validate it has the minimum required permissions and warn if not.

TDD REQUIRED: (1) Write a failing unit test that asserts a warning is logged when the role lacks required policies. (2) Implement the check. (3) make test must pass. (4) This changes command output (new warning) — add e2e test if feasible. Single-issue PR.

Files: client/src/utils/. Tests: tests/client/.

## Implementation notes

Implemented: validate custom --XRole permissions (trust policy + managed policies) and log warnings. TDD: 13 new unit tests, full suite green (1336 pass).

## Refinery handoff

- Issue: `bu-iot` (task, P2)
- Source branch: `polecat/bu-iot`
- Target: `main`

### DIFF
--- a/tools/bulk_executor/client/src/utils/role_validator.py
+++ b/tools/bulk_executor/client/src/utils/role_validator.py
@@ -1,0 +1,93 @@
+"""Validate minimum permissions on a custom IAM role for the Bulk Executor Glue job.
+
+Returns a list of human-readable warning strings. Empty list means all checks pass.
+Never raises or exits — validation failures are advisory only.
+"""
+from __future__ import annotations
+
+from utils.logger import log
+
+GLUE_SERVICE_PRINCIPAL = "glue.amazonaws.com"
+
+REQUIRED_MANAGED_POLICY_ARNS = [
+    "arn:aws:iam::aws:policy/service-role/AWSGlueServiceRole",
+]
+
+DYNAMODB_POLICY_ARNS = [
+    "arn:aws:iam::aws:policy/AmazonDynamoDBReadOnlyAccess",
+    "arn:aws:iam::aws:policy/AmazonDynamoDBFullAccess",
+]
+
+
+def validate_custom_role_permissions(
+    iam_client, role_name: str
+) -> list[str]:
+    """Check that a custom role has the minimum permissions for the Glue job.
+
+    Returns a list of warning messages (empty if role looks good).
+    Gracefully returns [] on any IAM API failure.
+    """
+    warnings: list[str] = []
+
+    try:
+        role_resp = iam_client.get_role(RoleName=role_name)
+    except Exception:
+        log.debug(f"Could not inspect role '{role_name}' — skipping permission validation")
+        return []
+
+    try:
+        policies_resp = iam_client.list_attached_role_policies(RoleName=role_name)
+    except Exception:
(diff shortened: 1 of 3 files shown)


hunterhacker:
Not a useful change. Here's what our README.md says about the Glue role. It's this list that needs to be checked:

If you provide a custom IAM role for your AWS Glue job:

- Ensure the role name starts with AWSGlueServiceRole
- Ensure the role has a trust policy that allows the Glue service principal (glue.amazonaws.com) to assume the role.
- Attach the managed policy AWSGlueServiceRole to grant Glue its baseline execution permissions.
- Attach ServiceQuotasReadOnlyAccess to allow the job to read service quota information (used to detect account-level read/write limits), or for maximum lockdown allow the pricing:GetProducts action.
- Attach AWSPriceListServiceFullAccess to allow the job to query AWS pricing APIs (used to estimate DynamoDB operation costs), or for maximum lockdown allow the servicequotas:GetServiceQuota and servicequotas:GetAWSDefaultServiceQuota actions.
- Add custom IAM permissions for DynamoDB access. You may attach AmazonDynamoDBReadOnlyAccess or AmazonDynamoDBFullAccess, or define a more restrictive policy targeting specific tables.

relentlesscol:
Reworked this to address the feedback — the original only matched attached-policy ARNs and checked a couple of items, which is why it wasn't useful. It now checks every requirement in the README list, and crucially checks *effective permissions* rather than matching managed-policy ARNs, so the documented "maximum lockdown" alternatives are accepted instead of false-warned:

- Role name starts with `AWSGlueServiceRole`
- Trust policy allows `glue.amazonaws.com`
- `AWSGlueServiceRole` managed policy attached
- Service Quotas read access — via `ServiceQuotasReadOnlyAccess` **or** the `servicequotas:GetServiceQuota` / `GetAWSDefaultServiceQuota` actions granted inline
- Pricing read access — via `AWSPriceListServiceFullAccess` **or** `pricing:GetProducts` granted inline
- DynamoDB access — via a managed policy **or** a restrictive per-table inline policy

It collects the actions the role effectively allows (from attached managed-policy documents plus inline policies) and checks the capabilities are present however they were granted. Advisory only: any IAM read failure degrades to skipping that check and never blocks bootstrap.

Added a live e2e suite (`test-e2e-security-validator`) that builds throwaway roles at known shapes against real IAM and asserts the validator warns exactly when it should — including the false-green guards: a fully-valid lockdown role produces zero warnings, and a restrictive inline DynamoDB policy does not false-warn. Unit tests rewritten to the effective-permission contract (100% line + branch on the validator).

tools/bulk_executor/client/src/utils/role_validator.py
```diff
@@ -0,0 +1,256 @@
+"""Validate minimum permissions on a custom IAM role for the Bulk Executor Glue job.
+
+Checks a custom ``--XRole`` against the requirements documented in README.md
+under "If you provide a custom IAM role for your AWS Glue job". Returns a list
+of human-readable warning strings; an empty list means every documented
+requirement is satisfied.
+
+Advisory only: this never raises and never exits. Any IAM read that fails
```
relentlesscol: Do we want advisory here?

PR #231, opened 2026-07-02
[bulk][split-217 #89] Expand rate warnings for too-high/too-low values (bu-9w5)

relentlesscol:
## Summary

Issue #89. Warn when the requested or table-derived read/write rate is a poor fit for the table — too **high** for what the table can deliver, too **low** to finish in the time remaining before the job times out — so the job doesn't silently throttle against a ceiling the caller can't raise, or run past the Glue timeout without warning.

This is a **rework** to Jason's 5-point spec (see review thread). The first attempt only read the title of #89. It now implements **all five checks** (#1–#5) and expands the permission/degradation handling from review feedback.

## What it does

Billing-mode-aware capacity checks in `get_dynamodb_throughput_configs`:

- **Provisioned, no autoscaling, request > provisioned** → hard warn (we can't raise it) *(spec #2)*
- **Provisioned, autoscaling on, request > autoscaling max** → hard warn *(spec #3)*
- **Provisioned < request ≤ autoscaling max** → softer scale-up note
- **On-demand, request > table `OnDemandThroughput` max** → hard warn *(spec #4)*
- **On-demand, no table max, request > account quota** → hard warn *(spec #5)*
- **Effective rate too slow to move the table's data in the time remaining before the job timeout** → "will likely time out" warn *(spec #1)*

Reuses the existing application-autoscaling + `get_quota_value` plumbing. Warnings are observational only — the return value is unchanged. Uses `log.warning`.

## Spec #1 — the timeout/ETA warning, now against time *remaining* (review follow-up)

Check #1 is a **planning-time estimate**: "look at the table metadata, decide how long the rate will take, and if it exceeds the budget, warn." Every input already lives in `get_dynamodb_throughput_configs` — the effective rate (user-set or table-derived), `TableSizeBytes`/`ItemCount` from the `describe_table`, and `--XTimeout` — so it sits alongside checks #2–#5 in `table_info`.

**The fix in Jason's latest review:** the budget must be the time **remaining** before the timeout, not the raw timeout. Multi-phase verbs run sequentially — `delete` does a scan phase, then a write phase — and resolve the *later* phase's rate only after an earlier phase has already burned part of the timeout. Checking a delete's write phase against the full 60 min is wrong when the scan already consumed 45 of them; it has to be checked against the ~15 min left.

Implementation:

- A monotonic reference is captured at module import (which on a Glue worker is job startup — `root.py` imports the verb, which imports `table_info`). `_job_elapsed_minutes()` returns wall-clock elapsed since then, clamped non-negative.
- `get_dynamodb_throughput_configs` computes `remaining = timeout − elapsed` and passes that to `_warn_if_job_may_timeout`, whose message now names the **time remaining** rather than the raw timeout.
- Because this function is called **when each phase begins**, a late phase automatically sees a shrunken budget — no per-verb plumbing, and no cross-phase state. This deliberately does **not** predict what future phases will need (per the spec).
- **No new IAM**: elapsed comes from the local monotonic clock, not `glue:GetJobRun`, so the permission envelope is unchanged.

I confirmed no verb calls `get_dynamodb_throughput_configs` earlier than the phase that needs it (find's delete resolves the write rate *after* the scan `count()`; reads resolve at connector `load()` time), matching Jason's note that no verb calls it overly early.

`_warn_if_job_may_timeout` reuses the same unit formulas as the cost estimators (read = `ceil(size/8096)`; write = `item_count * ceil(avg_item_size/1024)`), divides by the effective rate, and warns if the estimate exceeds the remaining budget. Best-effort: missing/zero metadata or a non-numeric rate simply skips.

## Permissions envelope (review follow-up)

The autoscaling-aware checks read a provisioned table's autoscaling maximum, which the generated Glue role could not previously see. This PR expands the permission envelope, and does all three things that entails:

- **Bootstrap:** new `MinimalAutoScalingAccess` inline policy granting `application-autoscaling:DescribeScalableTargets` on `Resource: "*"` (the action has no resource-level scoping).
- **README:** documents the new permission in the custom-role list, including the degradation behavior.
- **Version bump** `1` → `2`: existing bootstraps pick up the new policy via the version-mismatch role refresh (#84/#233).

**Graceful degradation (Jason's option b):** if the role lacks the permission, the job does **not** error — it skips the autoscaling-aware capacity check and logs a `log.warning` that it's proceeding without visibility into the table's autoscaling metrics, naming the missing permission. No false "exceeds provisioned" is emitted. The permission stays genuinely optional; granting it just re-enables the check.

## Second call site found live — and fixed

A live e2e run surfaced a bug this PR now fixes. There are **two** places that read a provisioned table's autoscaling settings, and graceful degradation has to cover both:

1. `_effective_capacity_ceiling` — the capacity-warning path. Already caught the AccessDenied and degraded.
2. `get_and_print_dynamodb_table_info` — the "Auto Scaling Settings" **diagnostic print** that runs *earlier* (`load/__init__.py` calls it before the write). This `describe_scalable_targets` was **unguarded** (pre-existing, commit `5a933d1`), so a provisioned `load` with the permission denied crashed there at info-print time, **before any data moved**. The traceback pointed at `load/__init__.py:117` (the `except` re-raise) because `from None` masked the real origin — which made it look like a connector-write failure; it was not.

Fix: wrap the diagnostic describe loop in `try/except`. On denial it logs `"Could not read autoscaling settings … skipping this diagnostic. Grant application-autoscaling:DescribeScalableTargets …"` and proceeds. With both call sites guarded, a provisioned load whose role is denied the permission runs to **SUCCEEDED**.

## Tests

TDD — `tests/server/test_rate_capacity_warnings.py` (all five checks). Check #1 now has coverage for the remaining-budget rework: `_job_elapsed_minutes` (zero at start, grows with the clock, clamped non-negative), `_warn_if_job_may_timeout` racing the time *remaining* rather than the raw timeout (a rate that fits a full 60-min timeout still warns when only 10 min remain), and an `elapsed-shrinks-the-budget` wiring regression through `get_dynamodb_throughput_configs` (mimics the multi-phase delete: the write rate resolves after ~55 min elapsed, leaving ~5 min for an ~8-min job → warns; early in the job → quiet). Plus `test_table_info.py::test_autoscaling_describe_denied_does_not_crash_print` (the diagnostic-print guard) and bootstrap-policy assertions in `tests/client/test_bootstrap.py`. `table_info.py` is at 100% line+branch coverage. Full offline suite: **1421 passing**, line 94.6%, branch 90.6%. Badges updated.

### E2E (opt-in, real Glue — not in `make test`)

`make test-e2e-capacity-warnings` proves the behavior on **real Glue jobs** — full suite **6 passed** (5 whole-system + 1 security):

- `tests/e2e/whole_system/test_capacity_warnings.py` — the four capacity warnings surface against real tables of each billing topology (provisioned-exceeds / autoscaling-max / soft scale-up / on-demand-max), plus **check #1**: a slow `--XMaxWriteRate` into the persistent, already-large `write_table` makes the estimate predict a multi-hour run and warn, while only ~20 rows are actually written so the job SUCCEEDS in ~2 min. Loading (not reading) decouples the estimate — keyed off the target's existing `ItemCount` — from the real work (the tiny input CSV); a read can't decouple these, since throttling to trip the estimate also throttles the real scan. Guards against a false-green (asserts the target's real `ItemCount` forces the estimate past the timeout before trusting the pass).
- `tests/e2e/security/test_capacity_warning_missing_perm.py` — a provisioned load whose Glue role is **explicitly denied** the permission emits both degradation notes **and the job reaches SUCCEEDED**. `CommandResult.output` (stdout+stderr) was added because Glue LiveTail routes `log.warning` to stderr.

**Note on the remaining-budget rework:** this suite's check-#1 scenario is a single-phase `load`, so elapsed ≈ 0 and remaining ≈ the raw timeout — it proves the warning still surfaces live, and the message substring it asserts on is unchanged. The specific late-phase "budget shrank because a scan already ran" behavior is proven by the unit regression above; a bespoke live multi-phase delete proof would need a table large enough for the scan to burn real minutes, which the unit test covers deterministically. The live suite runs on the fork's `fork-ci` pipeline as a background sanity check.

`transient_table` was extended for PROVISIONED / autoscaling-target / on-demand `MaxWriteRequestUnits` shapes.

## Rebase note

Rebased onto current `main` — includes #234 (issue #182) and #236/#237 (connector read-rate resolution). #236/#237 touch `glue_connector.py`, not the `table_info` capacity path, so no logic overlap; clean rebase.

### DIFF
--- a/tools/bulk_executor/Makefile
+++ b/tools/bulk_executor/Makefile
@@ -1,5 +1,5 @@
 .DEFAULT_GOAL := help
-.PHONY: help install test test-client test-server coverage clean test-e2e test-e2e-connector test-e2e-cleanup test-e2e-security test-e2e-security-simulator test-e2e-security-real test-e2e-commands test-e2e-whole-system test-e2e-max-rate
+.PHONY: help install test test-client test-server coverage clean test-e2e test-e2e-connector test-e2e-cleanup test-e2e-security test-e2e-security-simulator test-e2e-security-real test-e2e-commands test-e2e-whole-system test-e2e-max-rate test-e2e-capacity-warnings
 
 VENV := .venv
 PIP := $(VENV)/bin/pip
@@ -27,6 +27,7 @@ help:
 	@echo "  make test-e2e-commands              Bulk-command smokes against transient tables"
 	@echo "  make test-e2e-whole-system          True e2e: 60k load round-trip + observed rate enforcement"
 	@echo "  make test-e2e-max-rate              EXPENSIVE: prove load beats the old 60k connector ceiling (~\$$ per run)"
+	@echo "  make test-e2e-capacity-warnings     #89 rate-vs-capacity warnings, proven live (~5 short jobs)"
 	@echo "  make test-e2e-cleanup               Delete orphaned items left by a crashed e2e run"
 
 install:
@@ -114,6 +115,19 @@ test-e2e-max-rate:
 	  --override-ini="testpaths=tests/e2e" \
 	  -m "max_rate" \
 	  --e2e-suite="max-rate (beats legacy 60k ceiling)"
+
+# Issue #89 capacity/rate warnings, proven LIVE. Runs the four transient-table
+# scenarios (whole_system) and the missing-permission degradation (security)
+# serially in one process. Serial is required: the missing-perm test flips the
+# shared job's role and restores it in a finally, so it must not overlap the
+# write-capable scenarios. Each warning fires at throughput-config setup, so
+# fixtures are tiny — cost is ~5 short Glue jobs, not a max-rate run.
+test-e2e-capacity-warnings:
+	$(PYTEST) tests/e2e/whole_system/test_capacity_warnings.py \
+	  tests/e2e/security/test_capacity_warning_missing_perm.py \
+	  -s -v --tb=short --no-cov \
+	  --override-ini="testpaths=tests/e2e" \
+	  --e2e-suite="capacity warnings (#89, live)"
 
 test-e2e-cleanup:
 	$(VENV)/bin/python -m tests.e2e.helpers.cleanup
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -1,8 +1,8 @@
 # Bulk Executor for Amazon DynamoDB
 
-![tests](https://img.shields.io/badge/tests-1392%20passing-brightgreen)
+![tests](https://img.shields.io/badge/tests-1395%20passing-brightgreen)
 ![line coverage](https://img.shields.io/badge/line%20coverage-94.6%25-brightgreen)
-![branch coverage](https://img.shields.io/badge/branch%20coverage-90.5%25-brightgreen)
+![branch coverage](https://img.shields.io/badge/branch%20coverage-90.4%25-brightgreen)
 
 Bulk Executor for Amazon DynamoDB lets you efficiently run bulk commands against even large tables. It:
 
@@ -327,6 +327,7 @@ If you provide a custom IAM role for your AWS Glue job:
 * Attach the managed policy `AWSGlueServiceRole` to grant Glue its baseline execution permissions.
 * Attach `ServiceQuotasReadOnlyAccess` to allow the job to read service quota information (used to detect account-level read/write limits), or for maximum lockdown allow the `pricing:GetProducts` action.
 * Attach `AWSPriceListServiceFullAccess` to allow the job to query AWS pricing APIs (used to estimate DynamoDB operation costs), or for maximum lockdown allow the `servicequotas:GetServiceQuota` and `servicequotas:GetAWSDefaultServiceQuota` actions.
+* Allow the `application-autoscaling:DescribeScalableTargets` action (used to detect a provisioned table's autoscaling maximum when warning that a requested rate exceeds the table's capacity). This action does not support resource-level scoping, so it must be granted on `"Resource": "*"`. If the role lacks this permission the job still runs — it simply skips the autoscaling-aware capacity warning and logs that it is proceeding without visibility into the table's autoscaling settings.
 * Add custom IAM permissions for DynamoDB access. You may attach `AmazonDynamoDBReadOnlyAccess` or `AmazonDynamoDBFullAccess`, or define a more restrictive policy targeting specific tables.
 
 ### Security: Consider adjusting S3 bucket behaviors
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -1,8 +1,8 @@
 # Bulk Executor for Amazon DynamoDB
 
-![tests](https://img.shields.io/badge/tests-1330%20passing-brightgreen)
-![line coverage](https://img.shields.io/badge/line%20coverage-94.1%25-brightgreen)
-![branch coverage](https://img.shields.io/badge/branch%20coverage-91.7%25-brightgreen)
+![tests](https://img.shields.io/badge/tests-1392%20passing-brightgreen)
+![line coverage](https://img.shields.io/badge/line%20coverage-94.6%25-brightgreen)
+![branch coverage](https://img.shields.io/badge/branch%20coverage-90.5%25-brightgreen)
 
 Bulk Executor for Amazon DynamoDB lets you efficiently run bulk commands against even large tables. It:
 
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -1,6 +1,6 @@
 # Bulk Executor for Amazon DynamoDB
 
-![tests](https://img.shields.io/badge/tests-1395%20passing-brightgreen)
+![tests](https://img.shields.io/badge/tests-1396%20passing-brightgreen)
 ![line coverage](https://img.shields.io/badge/line%20coverage-94.6%25-brightgreen)
 ![branch coverage](https://img.shields.io/badge/branch%20coverage-90.4%25-brightgreen)
 
@@ -626,6 +626,7 @@ The e2e harness has these suites:
 | Security  | `make test-e2e-security`  | the documented bootstrap IAM policy actually bootstraps (and is minimal) |
 | Whole-system | `make test-e2e-whole-system` | true end-to-end: 60k `load` round-trip fidelity + observed write-rate enforcement from CloudWatch |
 | Max-rate | `make test-e2e-max-rate` | **expensive, opt-in:** proves `load` sustains a write rate above the old connector's 60k WCU/s ceiling (millions of items + pre-warmed table) |
+| Capacity warnings | `make test-e2e-capacity-warnings` | issue #89: `load --XMaxWriteRate` above a table's ceiling emits the right warning live (provisioned / autoscaling-max / autoscaling-soft-note / on-demand-max), plus the missing-`DescribeScalableTargets` visibility degradation — asserted in the real Glue job's log stream |
 
 Each command/connector smoke creates its own short-lived table (`bulk-e2e-<command>-<random>`, tagged `ephemeral=true`) and **tears it down in a `finally` block** even on failure — the suite never touches your existing tables. If a run is hard-killed mid-test, sweep any orphans with `make test-e2e-cleanup`. The first e2e run prompts once for account/region/test-table config and caches it in `tests/e2e/.e2e-config` (gitignored, per-developer).
 
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -1,8 +1,8 @@
 # Bulk Executor for Amazon DynamoDB
 
-![tests](https://img.shields.io/badge/tests-1396%20passing-brightgreen)
+![tests](https://img.shields.io/badge/tests-1412%20passing-brightgreen)
 ![line coverage](https://img.shields.io/badge/line%20coverage-94.6%25-brightgreen)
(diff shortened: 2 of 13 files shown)


hunterhacker:
With this PR it seems like the AI read the title of #89 but not the body text explaining what checks we want.

relentlesscol:
Merge-order note: this PR edits the same `get_dynamodb_throughput_configs` region as #234 (issue #182), which removes the old percentage-cap logic. #234 is verified and should merge first; please rebase this onto the resulting `main` and re-resolve the positional overlap. The changes are conceptually independent — #234 removes dead percentage narration, this adds net-new too-high/too-low capacity warnings — so this should be a mechanical rebase, not a logic merge.

hunterhacker:
This fix doesn't follow the logic laid out in #89 

1. Print a warning if the next big part of a job is likely to take longer than the remaining time before timeout. Look at the timeout and the time elapsed so far. Look at the read and write rates. Look at the expected read and write consumption of the next big task going on. If the needed time is within some nearby threshold of the time remaining (like will take more than 90% of what's remaining), give a warning. This is like the nice feature when you navigate your car to a restaurant and your car points out it'll likely be closed by the time you get there. For tasks that do actions in sequence like deletes where there's a read phase and write phase. The read phase should look at the time remaining and not worry about the delete phase. Then the delete phase should look not at the time remaining on its own. When it's the first task the time remaining is basically the timeout but since some commands are multi-phase we should look at time remaining not timeout. For now let's not try to predict what future phases will require.

2. Warn if the table is provisioned and doesn't have autoscaling on and is hard-provisioned below their requested rate. Cuz we can't raise it up.

3. Warn if the table does have autoscaling but the max autoscaling setting is below their requested rate.

4. Warn if the table is on-demand but the max setting on the specific table is below their requested rate.

5. Warn if the table is on-demand and there's no max setting on the table but the account quota is below their requested rate.

Later we can have warnings as well if the new consumption will put them over the limits or allowances.

relentlesscol:
Reworked to your 5-point spec, @hunterhacker — you're right that the first pass followed the title of #89 and not the body. I threw out the single "exceeds capacity" comparison and the off-spec "< 1% too-low" check and rebuilt it around your list:

- **#2** provisioned, no autoscaling, hard-provisioned below the requested rate → warn (can't raise it) ✅
- **#3** provisioned with autoscaling, max below the requested rate → warn ✅
- **#4** on-demand, table's own max below the requested rate → warn ✅
- **#5** on-demand, no table max, account quota below the requested rate → warn ✅

Plus a softer scale-up note for the in-between case (provisioned floor < request ≤ autoscaling max). All billing-mode aware, warnings are observational only, and the autoscaling-lookup-failure path is skipped so we don't emit a false "exceeds provisioned". 35 new tests, TDD.

**#1 (the timeout/ETA "restaurant will be closed" warning) is deferred to a follow-up PR** — it needs time-remaining and per-phase read/write consumption context that lives in the runner, not in `table_info` where these capacity checks sit. I kept your framing in mind (look at time *remaining* not the raw timeout, treat multi-phase read/write phases independently, don't predict future phases yet) so the follow-up drops in cleanly.

Also rebased onto `main` now that #234 (issue #182) has merged — that removed the old percentage-cap narration from the same `get_dynamodb_throughput_configs` region, so this sits cleanly on top as net-new warnings. My earlier merge-order note is now resolved.

hunterhacker (commented):
Code looks elegant. Looking forward to e2e testing of the various code paths because there's a lot of sharp edges in how it pulls in the metrics.

Let's also make sure our generated Glue roles have the right permissions to do all this new reading by default and we've documented the right permissions in the README, if any new ones are needed.

hunterhacker:
Think the Glue role would need `application-autoscaling:DescribeScalableTargets` added. That's a change in bootstrap making the roles and in the README if you make a custom role. And needs a bump in the version!

If it's not there we can (a) error out or (b) just proceed without being smart about the warning. If we did (b) we should then warn that we can't see the autoscaling targets so proceeding without knowledge of the table's metrics.

relentlesscol:
Good catch on the permissions envelope, @hunterhacker — you're right, the autoscaling reads were load-bearing without the role being able to see them. Handled all three parts plus the fallback:

- **Bootstrap** now attaches a `MinimalAutoScalingAccess` inline policy granting `application-autoscaling:DescribeScalableTargets`. That action has no resource-level scoping, so it's on `Resource: "*"` (noted in code + README).
- **README** custom-role list documents the new permission and the degradation behavior.
- **Version bumped 1 → 2**, so existing bootstraps pick the policy up automatically through the version-mismatch role refresh from #84/#233 — no manual re-grant needed.
- **Fallback = your option (b):** if the permission is absent we don't error. We skip the autoscaling-aware check and emit a `log.warning` that we're proceeding without visibility into the table's autoscaling metrics, naming the missing permission so it's actionable. No false "exceeds provisioned".

Also rebased onto this morning's `main` (picks up #236/#237). Those touch `glue_connector.py`, not the `table_info` capacity path, so it was a clean rebase with no logic overlap.

On e2e: agreed there are sharp edges in how the metrics get pulled (autoscaling on/off, provisioned vs on-demand, table max vs account quota, and now the missing-permission path). I'll wire these scenarios into the e2e connector suite against transient tables before we call this done — unit coverage alone won't prove the metric-reading paths.

relentlesscol:
Good for review, @hunterhacker. 🟢

All five #89 checks are now implemented and the diagnostic-print guard is in. Check #1 (the timeout/ETA warning) landed here rather than as a follow-up — it turned out to be a planning-time estimate that reuses the table metadata already read in `get_dynamodb_throughput_configs`, so it sits alongside checks #2–#5 in `table_info` with no runner/phase plumbing needed.

**Vetted end-to-end before flagging ready:**

- **Offline:** `make test` = 1412 passing, line 94.6% / branch 90.6%; `table_info.py` at 100% line+branch.
- **Live capacity-warnings e2e** (`make test-e2e-capacity-warnings`, real Glue on us-east-1): 6 passed — all five warnings fire on real jobs of each billing topology, plus the missing-permission graceful degradation. Check #1 proved live: a slow `--XMaxWriteRate` into a 2,000,004-item table logged `Estimated write time of ~333 min … will likely time out` while only ~20 rows were written, so the job still SUCCEEDED (~2 min).
- **Full fork CI green** (unit → connector-smoke → command-e2e, real Glue/DynamoDB): https://github.com/relentlesscol/amazon-dynamodb-tools/actions/runs/29441345097

Still a single, self-contained PR — capacity/timing warnings in `table_info`, observational only, return value unchanged.

hunterhacker:
Change looks pretty surgical and appropriate. 

But I don't see how this part of my comment here was addressed: "For tasks that do actions in sequence like deletes where there's a read phase and write phase. The read phase should look at the time remaining and not worry about the delete phase. Then the delete phase should look not at the time remaining on its own. When it's the first task the time remaining is basically the timeout but since some commands are multi-phase we should look at time remaining not timeout. For now let's not try to predict what future phases will require."

If we look at the code in `find.py` that handles deletes, the write throughput gets calculated after the scan finishes and that point might be 45 minutes into a 60 minute timeout. It's not right to check if the delete will be done in 60 minutes, it has to be in the 15 minutes remaining. So the math to figure out time remaining should be the timeout but the time remaining (timeout minus the elapsed time since the start of the Glue job).

Then it's up to the code in each verb to not call `get_dynamodb_throughput_configs` before it's needed. I think no verbs call overly early.

relentlesscol:
Good catch, @hunterhacker — you're right, I'd wired check #1 to the raw timeout, not the time remaining. Fixed at `0688b02`.

The budget is now `timeout − elapsed`, measured at the instant each phase resolves its rate:

- A monotonic reference is captured at module import (job startup on the worker). `_job_elapsed_minutes()` returns wall-clock elapsed since then.
- `get_dynamodb_throughput_configs` computes `remaining = timeout − elapsed` and the timeout warning races against that. The message now names the time remaining, not the timeout.
- Because the function is called **when each phase begins**, a late phase sees a shrunken budget for free — no cross-phase state, no per-verb plumbing. Your `find.py` delete example is exactly the case: the scan runs first, then the write rate resolves ~45 min in, so the write phase is now checked against the ~15 min left, not the full 60. Per your steer, it looks at time remaining and doesn't try to predict future phases.
- No new IAM — elapsed is the local monotonic clock, not `glue:GetJobRun`, so the permission envelope is unchanged.

I also confirmed your other point: no verb calls `get_dynamodb_throughput_configs` earlier than the phase that needs it (delete resolves the write rate after the scan `count()`; reads resolve at connector `load()` time).

Tests: added `_job_elapsed_minutes` coverage, remaining-vs-raw-timeout cases (a rate that fits a full 60-min timeout still warns when only 10 min remain), and an elapsed-shrinks-the-budget wiring regression that mimics the multi-phase delete. `make test`: 1421 passing, line 94.6% / branch 90.6%; `table_info.py` still 100% line+branch.

On e2e: the live check-#1 scenario is single-phase `load` (elapsed ≈ 0), so it proves the warning still surfaces but not the late-phase shrink — that's covered by the unit regression, and the live suite is running on the fork CI as a background sanity check. Not merging on blog go-live day; this is staged and ready.

hunterhacker:
Code looks good. I'll experiment with it next.

hunterhacker:
Checking the output it says: 

```
Requested read rate 110000 exceeds the table's table's on-demand maximum of 100000; the table cannot deliver this rate
```

Notice the double `table's`.

It's funky because what it's saying is it exceeds the table's thing called a "table's on-demand maximum".

Perhaps it should say:

```
Requested read rate 110000 exceeds the table's on-demand maximum throughput of 100000; the table cannot deliver this rate
```

PR #233, opened 2026-07-13
fix(bootstrap): refresh built-in Glue role on version mismatch (#84)

### LINKED ISSUE #84: Force built-in role refresh on version mismatch

relentlesscol:
Closes #84.

## Problem
`bootstrap` never overwrites the generated Glue role. When a newer version of bulk_executor adds permissions to that role, a user who bootstrapped on an older version is left with a role missing required permissions -- failing in subtle ways (e.g. can't write to CloudWatch, so no output). Today's only workaround is a full teardown/bootstrap.

## Fix
Detect a version mismatch (deployed Glue job's `--bulk-dynamodb-version` vs local `__version__`) and, when the role already exists, refresh it instead of returning early:

- Re-apply the attached/inline policies (as before), **and**
- Re-apply the **trust policy** via `update_assume_role_policy`.

That second part matters: `create_role` (which sets the trust policy) is skipped for an existing role, so without this a `trust_policy` change + version bump would leave a refreshed role with a stale `AssumeRolePolicyDocument`. Now a refreshed role is identical to one a fresh bootstrap would create.

Emits a clear warning naming both versions and the resolution (redeploy the Glue job).

## Tests
Added `test_bootstrap.py` coverage including a test asserting `update_assume_role_policy` is called on refresh with the `glue.amazonaws.com` trust principal. Full bootstrap suite: 75 passed.

## Note
This replaces #230, which carried unrelated `rate_limiter` test files from a stale branch base and a noisy review thread. This PR is the same fix rebased clean onto `main` -- 2 files, +169/-1.


relentlesscol:
Added a real-AWS e2e test (\`tests/e2e/security/test_real_role_refresh.py\`, opt-in via \`make test-e2e-security-real\`) so this is verified end-to-end, not just against mocks.

It snapshots the shared Glue job's role trust policy + deployed version, corrupts both (injects a bogus \`ec2\` principal into the trust policy + sets \`--bulk-dynamodb-version\` to an older value), runs real \`./bulk bootstrap\`, then asserts via a real \`iam.get_role\` that the trust policy converged back to the glue-only shape a fresh bootstrap produces. Restores original state in \`finally\`, pass or fail.

Verified passing against a real account:
```
WARNING Version mismatch: deployed Glue job is v0, but you are running v1 locally...
...
tests/e2e/security/test_real_role_refresh.py::test_version_mismatch_refresh_converges_trust_policy_real_iam PASSED
1 passed in 9.89s
```
Unit suite (75) still green.

relentlesscol:
Updated the e2e test to be fully isolated: it now creates its own throwaway role, corrupts + refreshes THAT, and deletes it in \`finally\` -- it no longer touches the shared \`bulk_dynamodb\` job or its role. So it is safe to run under parallel test runs and while a live Glue job is running (the shared role is never mutated). Still verified passing against real IAM (v0-vs-v1 mismatch warning fires, trust policy converges to glue-only), confirmed no leaked roles afterward.

hunterhacker:
I don't understand the warning here:

```
def _needs_role_refresh(self):
        job_details = self._get_glue_job_details()
        if not job_details:
            return True
        deployed_version = job_details['Job']['DefaultArguments'].get('--bulk-dynamodb-version')
        if not deployed_version:
            return True
        if deployed_version != VERSION:
            log.warning(
                f"Version mismatch: deployed Glue job is v{deployed_version}, "
                f"but you are running v{VERSION} locally. "
                f"IAM policies will be re-applied to ensure the role has permissions "
                f"required by v{VERSION}. To resolve this mismatch, redeploy the Glue job."
            )
            return True
        return False
```

The user is doing a bootstrap. If the versions don't match, that's probably why they're doing the bootstrap. We don't need to warn about it. We definitely don't need to say to redeploy in the middle of an active redeploy!

We could do an info level comment that we're refreshing the role permissions due to a version mismatch.

Please make sure this logic works if the client is an OLDER version as well as a NEWER version.

relentlesscol:
Good call, fixed. Three changes in ecb8e81:

- **Dropped to `info`.** A version mismatch during bootstrap is the expected reason someone's bootstrapping, so it's no longer a `warning`.
- **Removed the "redeploy the Glue job" line.** As you said, telling the user to redeploy in the middle of an active redeploy makes no sense.
- **Made it direction-agnostic.** The message no longer assumes local is newer, so it reads correctly whether the client is older or newer than the deployed job:

  > Version mismatch (deployed Glue job is v0, local is v1); refreshing the role's IAM policies to match the version being bootstrapped.

Updated the unit tests to match, plus added a test asserting it's never logged at `WARNING`+ and contains no directional (newer/older) wording. Full suite green (1255 passed).

PR #234, opened 2026-07-13
fix(load): enforce write rate via connector + report how it was determined (#182)

### LINKED ISSUE #182: The "load" command makes no statements about write rates

relentlesscol:
Closes #182.

## Problem
`load` made no single, trustworthy statement about the write rate, and never truly enforced one. A user could pass `--XMaxWriteRate`, but the connector write path turned it into a `dynamodb.throughput.write.percent` (with a 1.5x cap and an on-demand denominator guess), so the rate the job actually used was opaque and often not what was asked for. And even once the rate *was* wired through correctly, `load` could not deliver it: the write was pinned to `repartition(30)`, so a large `--XMaxWriteRate` was capped by Spark parallelism, not by the value the user set.

### 1. Wire the rate through as a direct integer (no percentage math)
Matching the Glue 5.x connector's native support for absolute rates:

- `load/run()` calls `get_dynamodb_throughput_configs(modes=["write"], format="connector")` to resolve the effective write rate (user `--XMaxWriteRate` > table-level > quota > default 40k) and passes it as an explicit `write_rate` kwarg.
- `glue_connector.write_dynamodb_dataframe` accepts `write_rate`, which takes precedence over the `parsed_args` fallback, and sets `dynamodb.throughput.write` directly.
- `table_info.py` drops the percentage math (`.write.percent`, 1.5x cap, 40k denominator) and emits direct `dynamodb.throughput.write` / `dynamodb.throughput.read` integers. Also fixes two latent `int(None)` `TypeError`s when a provisioned table reports no RCU/WCU.

### 2. Report the rate in one place, server-side
The resolved rate — **and how it was determined** — is stated by the worker during the run (`table_info.py`), e.g. `Max write rate set to specified limit / account quota / default on-demand / table-specific`. A rate should be reported once, wherever it is actually resolved, regardless of how it was determined. (An earlier revision of this PR also echoed the user-specified rate client-side before dispatch; per review that was removed — it only fired when the user already typed the value, so it just echoed their own input back, and the server-side statement plus the CloudWatch-observed rate below are strictly stronger.)

### 3. Actually deliver the rate: repartition the write to 100 tasks
`load` collapsed the write to `repartition(30)` — 30 Spark tasks against a 220 G.1X worker cluster (~3.5% utilization). A 120k `--XMaxWriteRate` request stalled at ~86k observed WCU/s (72% of target): **parallelism, not the rate knob, was the ceiling.** Repartitioning to 100 makes the connector rate the binding limit again — the same 120k request now sustains ~118k WCU/s (99%) on real Glue (see the max-rate e2e below).

### Addresses all of #182 (original issue + follow-up review)
- **"load makes no statements about write rates"** — the resolved rate and its provenance are now stated server-side during the run.
- **"the passed-in rate isn't actually being used"** — the rate is enforced through the connector *and* the write path is parallelized enough to deliver it, both verified end-to-end below.
- **"use `table_info.py` / `get_dynamodb_throughput_configs` to determine the rate"** — that is exactly the call `load` now makes.
- **"fix the library to not do percentile any longer, just straight rates"** — the percentage approach is removed from the library.

### Unit
Connector/table_info/load coverage: `write_rate` override precedence, None-fallback, the rewritten connector contract (direct ints, no percent keys), and `load` repartitioning to 100 before calling the wrapper. A pinned connector test (`test_write_rate_above_legacy_60k_ceiling_passes_through`) asserts a rate above the old 60k ceiling passes straight through unclamped, kept in sync with the max-rate e2e. Full suite passes (1330).

### E2E — smoke (early detection)
`tests/e2e/connector/test_load_smoke.py`: runs `load --XMaxWriteRate` against real Glue, asserts `JobRunState==SUCCEEDED` and items landed. Proves the connector *accepts* the direct-int `dynamodb.throughput.write` option (a unit test can't — `./bulk` exits 0 even when the job FAILS).

### E2E — whole-system: round-trip + enforcement
`tests/e2e/whole_system/test_load_rate_roundtrip.py` drives a **60k-item load** against a transient table and asserts:
- **Round-trip fidelity** — all 60,000 items land, payload values intact on read-back.
- **Rate enforcement** — observed `ConsumedWriteCapacityUnits` (CloudWatch) peak minute stays within the requested **2000 WCU/s** ceiling.
- **No vacuous pass** — fails loudly if the load finished before the ceiling could bind.

### E2E — max-rate: the write path beats the legacy 60k ceiling
`tests/e2e/whole_system/test_load_exceeds_legacy_ceiling.py` (marked `max_rate`, **excluded from the default sweep** because it costs real money) proves the two fixes work together. It drives 18M tiny items across 150k partition keys into a table pre-warmed to 180k WCU/s and observes the sustained rate from CloudWatch — the unfakeable behavioral proof (the old 40k x 1.5 = 60k cap made an observed >60k physically impossible). Guards against a vacuous pass if warm throughput never provisioned. Run via `make test-e2e-max-rate`; tune with `BULK_E2E_MAXRATE_WCU` / `_ITEMS` / `_PARTITIONS`.

**Verified on 654654401288 / us-east-1** (branch code deployed to the Glue job):
- Smoke: live job runs SUCCEEDED, items landed.
- Whole-system round-trip: job SUCCEEDED, `--XMaxWriteRate=2000`, 60,000 items round-tripped, peak sustained WCU/s within ceiling.
- **Max-rate (2026-07-13):** `--XMaxWriteRate=120000`, job SUCCEEDED, 18M items written.
  - At `repartition(30)`: observed peak **86,058 WCU/s (72% of target)** — parallelism-bound.
  - At `repartition(100)`: observed peak **118,262 WCU/s (99% of target)** — knob-bound. Per-minute WCU/s `[…, 115867, 118262, 46483]`, zero DynamoDB throttling.

New helpers: `helpers/capacity.py` (CloudWatch WCU/s oracle), `WarmThroughput` support in `helpers/transient_table.py`, the `max_rate` marker, `make test-e2e-whole-system` / `make test-e2e-max-rate`, and a smoke-vs-true-e2e + suite-parallelism section in `tests/e2e/AGENTS.md`.

## Notes
- Supersedes #226 (only *reported* a rate it didn't enforce — @hunterhacker's review caught that) and #232 (auto-closed on branch rename; identical content). Both are closed.
- Also removes the read path's `dynamodb.throughput.read.percent` companion in favor of a direct `dynamodb.throughput.read` integer, so the connector read commands (count/find/sql) are re-verified via `make test-e2e-connector`.
- **Merge-order dependency:** #231 (issue #89, too-high/too-low rate warnings) edits the same `get_dynamodb_throughput_configs` region. It should merge *after* this PR and rebase onto the result. The two are conceptually independent — this PR removes dead percentage-cap narration; #231 adds net-new capacity-validation warnings.
- **Combined source + tests, by design:** the `repartition(100)` and connector changes are source shipped alongside their own tests, departing from the repo's tests-only / source-refactor split rule. The departure is intentional — the source change *is* the issue's remaining ask, not a coverage exercise, so it belongs in the PR that closes #182.
- **Follow-up (not in this PR):** `repartition(100)` is a hardcoded value giving a ~287k WCU/s ceiling on the current cluster; a `--XWritePartitions` knob can come later if a workload needs more. The read path's parallelism (`dynamodb.splits`) toward the same enterprise-scale target is a separate change.
- **Docs:** README badges synced to a full `make test` (1330 passing), and AGENTS.md instructions re-derive numbers from `make test` rather than bumping off a possibly-stale badge.
- **Merged `main` (2026-07-14):** resolved the one conflict @hunterhacker flagged, in `tests/e2e/helpers/transient_table.py`. Both sides were purely additive — `main` added an optional `pitr: bool = True` flag; this branch added `warm_write_units` / `warm_read_units` / `wait_for_warm`. Resolution is the union: PITR stays gated on `pitr`, warm-throughput provisioning and its wait are preserved. No behavioral change to either feature. Full suite green post-merge (**1357 passing**, 49 pre-existing documented skips), `mergeable: CLEAN`.


hunterhacker:
I don't like:

```
Client-side pre-dispatch statement (new):
```

This only happens when the user specified the rate so they don't need to see the rate they asked for printed back to them. Don't need more advanced if/then/else logic. Just output the rate in the same spot no matter how it's determined.

hunterhacker:
Haven't yet tested but code looks good now. AI, please resolve the conflicts.

relentlesscol:
@hunterhacker conflicts resolved and both review points addressed:

**Conflict** — the one file was `tests/e2e/helpers/transient_table.py`. Both sides were purely additive: `main` added an optional `pitr: bool = True` flag; this branch added `warm_write_units` / `warm_read_units` / `wait_for_warm`. Resolved as the union — PITR stays gated on `pitr`, warm-throughput provisioning + wait preserved, no behavioral change to either. `mergeable: CLEAN`, unit-tests CI green, full local suite 1357 passing.

**"Client-side pre-dispatch statement"** — removed. There is now no client-side rate echo at all; the rate is stated in exactly one place, server-side in `table_info.py`, on the same line regardless of how it was determined (specified limit / account quota / default on-demand / table-specific / provisioned). No if/then/else on the client. Ready for another look.

PR #238, opened 2026-07-15
ci(bulk_executor): bump checkout@v5 and setup-python@v6 for Node 24

relentlesscol:
## What

Bumps two GitHub Actions in `.github/workflows/bulk-executor-unit-tests.yml`:

- `actions/checkout@v4` → `@v5`
- `actions/setup-python@v5` → `@v6`

One-line-each change, no workflow logic or behavior change.

## Why

`checkout@v4` and `setup-python@v5` declare the **Node 20** runtime. GitHub is [deprecating Node 20 on Actions runners](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) and currently auto-forces these actions onto Node 24, emitting a deprecation annotation on **every** run of this workflow:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4, actions/setup-python@v5.

The `@v5`/`@v6` majors are the Node-24-native releases. Bumping clears the annotation before GitHub hard-removes Node 20 (at which point the pinned versions would fail rather than warn).

## Scope

Standalone CI hygiene — unrelated to any in-flight `bulk_executor` feature/test work. Touches only the workflow file; no source, tests, or docs.

## Verification

Ran the updated workflow on a fork's `main` before opening this PR:

- Job `unit-tests` → **success** (full offline suite, ~1m)
- Post-steps confirm the new pins (`Post Run actions/checkout@v5`)
- **Annotations: 0** — the Node 20 deprecation warning no longer appears

The workflow's own `paths` filter includes `.github/workflows/bulk-executor-unit-tests.yml`, so this change triggers the very workflow it modifies.

### DIFF
--- a/tools/bulk_executor/server/src/python_modules/shared/glue_connector.py
+++ b/tools/bulk_executor/server/src/python_modules/shared/glue_connector.py
@@ -9,6 +9,8 @@ attached to the job. Bootstrap handles both requirements; verbs go through
 this module without caring about the underlying Spark API.
 """
 
+from python_modules.shared.table_info import get_dynamodb_throughput_configs
+
 
 def read_dynamodb_dataframe(
     glue_context,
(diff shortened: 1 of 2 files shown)


relentlesscol:
Superseded — reopening from a dedicated branch (ci/node24-action-bumps) so the diff is strictly the two action bumps. This PR's head was the fork's main, which briefly carried a fork-only e2e workflow; closing to keep the upstream diff clean.

PR #240, opened 2026-07-16
[bulk_executor] fix(update): remove redundant TableName from scan_kwargs

relentlesscol:
## What

Removes the redundant `TableName` key from `scan_kwargs` in the `update` verb.

The verb builds `table = dynamodb_resource.Table(table_name)` and calls `table.scan(**scan_kwargs)`. On a Table **resource**, the scan is already bound to that table, so passing `TableName` is redundant (and an unexpected key on the resource-level API). This drops it while keeping `Segment` / `TotalSegments` for the parallel scan.

## Testing

- `test_scan_kwargs_include_segment_and_total` updated to assert `TableName` is absent. Verified RED→GREEN: reintroducing the key fails the test.
- Full offline suite green (`make test`): 1358 passing.

## Notes

Clean re-submission of the fix originally in #188, minus the unrelated `_jsonify_message` regex tests that PR had bundled in. No overlap with any other open PR.

### DIFF
--- a/tools/bulk_executor/server/src/python_modules/update/__init__.py
+++ b/tools/bulk_executor/server/src/python_modules/update/__init__.py
@@ -115,7 +115,6 @@ def _update_data(monitor_options, table_name, generate, segment, total_segments,
     skipped_count = 0
     failed_count = 0
     scan_kwargs = {
-        "TableName": table_name,
         "Segment": segment,
         "TotalSegments": total_segments
     }
--- a/tools/bulk_executor/tests/server/test_update.py
+++ b/tools/bulk_executor/tests/server/test_update.py
@@ -559,7 +559,7 @@ class TestUpdateDataPagination:
         updated_acc.add.assert_called_once_with(3)
 
     def test_scan_kwargs_include_segment_and_total(self, monkeypatch):
-        """Lines 117-121: scan_kwargs includes TableName, Segment, TotalSegments."""
+        """Lines 117-120: scan_kwargs includes Segment, TotalSegments (not TableName — the Table resource already knows its name)."""
         scan_kwargs_seen = []
         table = MagicMock()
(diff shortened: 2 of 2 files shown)


hunterhacker:
Looks reasonable. Will merge when Colin is back.

relentlesscol:
Ready to merge whenever you want to land it.

While re-verifying I noticed `scancount` had the same redundant `TableName` on a Table-resource scan, so I cleaned that up here too (`097db4c`) rather than leaving it for a follow-up. Title and description updated to match.

I swept for the rest of the pattern: those two were the only occurrences. `diff.py:62` keeps its `TableName` because that path scans through a low-level `client('dynamodb')` where the key is required, and the other six are `describe_table` calls where it's the actual API parameter. Also checked the neighbouring kwargs dicts for the same class of redundancy and found none.

Both tests now assert the key is absent instead of dropping the assertion, so a regression that reintroduces it fails.

Proven live rather than just green, since unit tests can't tell you boto3 accepts the kwargs against a real table: `scancount` over a 160,123,395-item table returned `Total records counted: 160,123,395`, matching `DescribeTable`'s ItemCount, job SUCCEEDED in 2:36. Offline suite 1358 passing.

Unrelated, but flagging two things I tripped over while verifying, both pre-existing and neither caused by this PR:

- `scancount` against a very small table (5 items) dies with `ValueError: Invalid rate limits: read=0.5, write=500. Both must be >= 1 capacity units per second.` Reproduces without this change. Looks like a rate floor that makes tiny tables unscannable. Happy to file an issue.
- The local/remote version gate (`Local is 1, remote is 2`) blocks a version-1 branch against an account bootstrapped by #231, which bumped to 2. Expected behaviour, just worth knowing it bites when testing older branches side by side.

PR #219, opened 2026-07-01
[bulk_executor] feat: add --XMaxEstimatedCostAllowed parameter

relentlesscol:
### Issues addressed
- #175 — Add a max cost parameter

---

Adds a `--XMaxEstimatedCostAllowed` parameter that gates job launch based on estimated cost, preventing accidentally expensive operations.

- Estimates cost from item count, rate, and pricing
- Blocks launch with clear error if estimate exceeds threshold
- Opt-in only (no default cap)

All tests pass.

PR #241, opened 2026-07-16
[bulk_executor] feat: add --XMaxEstimatedCostAllowed cost gate

relentlesscol:
## What

Adds an optional client-side **cost gate**: `--XMaxEstimatedCostAllowed`.

When set, the CLI estimates the DynamoDB cost of the operation before launching Glue (read cost for `--table`/`--source`, write cost for `--target`) using the Pricing API — with a hardcoded fallback if that call fails — and aborts with an actionable message if the estimate exceeds the threshold. Absent the flag, behavior is unchanged.

- `estimate_cost()`: per-table read/write cost; on-demand vs provisioned (1.5x discount); STANDARD vs Infrequent-Access table class; missing-table skip; pricing-API fallback.
- `check_cost_gate()`: no flag → pass-through; cost ≤ threshold → proceed; cost > threshold → `sys.exit` with guidance to raise the limit or reduce scope.
- Wired into `bulk` just before the Glue launch.

## Testing

- 14 unit tests (`test_cost_gate.py`), written RED-first (ImportError, functions absent) then GREEN. Covers IA pricing, zero-item-write skip, provisioned write discount, fallback pricing, and all three gate outcomes.
- New cost-gate code at 100% line coverage.
- Full offline suite green (`make test`): 1371 passing.

## Notes

Clean re-submission of #219, isolated from the unrelated `--XIdleTimeout` and error-handling changes its original stacked branch had bundled in. No file overlap with other open PRs.

Not yet exercised against the live Pricing API end-to-end — offline-verified only.

### DIFF
--- a/tools/bulk_executor/bulk
+++ b/tools/bulk_executor/bulk
@@ -93,6 +93,9 @@ if(is_client_and_server_action): # Only run the server if the client script DNE 
     args = utils.get_args_from_processed_args(processed_args)
     script_args = utils.convert_client_dict_to_script_args(processed_args)
 
+    # Cost gate - check estimated cost before launching Glue
+    utils.check_cost_gate(env_configs, processed_args)
+
     # Dev Mode - Push new code to S3 without a full bootstrap
     if(args.get('XDev')):
         BootstrapInfrastructure(env_configs).update_python_modules_in_s3()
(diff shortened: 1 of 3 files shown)


hunterhacker:
Not liking this change.

Our code already had cost estimation logic, within each command (verb). This PR creates a new isolated function to do cost estimates. I'll note that it's overly simplistic with hard-coded assumptions about what parameter names might be used and for what:

```
    if 'table' in processed_args:
        tables.append(('read', processed_args['table']))
    if 'source' in processed_args:
        tables.append(('read', processed_args['source']))
    if 'target' in processed_args:
        tables.append(('write', processed_args['target']))
```

Does it accurately do costing for `diff`? No.

Perhaps the right design would ask the command/verb for its cost estimate. So then `diff` can do the read costs on both tables, for example. And while doing that we can make the way each verb does estimates a little more elegant.

Note that `delete` has two phases so that's tricky.

relentlesscol:
Closing PR to reset agent baseline.

PR #242, opened 2026-07-16
[bulk_executor] feat(scancount): --per-segment skew report and --segments control

relentlesscol:
## What

Adds two related `scancount` options, addressing the review feedback on #190:

- **`--per-segment`**: after the total, prints each scan segment's item count (sorted descending, with % of total), a mean, and a **skew ratio** (hottest segment / mean). Warns when the ratio exceeds 5x — surfacing hot partitions / uneven key distribution.
- **`--segments`**: controls the parallel scan segment count (default 200, matching the prior hard-coded value). Requested in the #190 review; also improves per-segment resolution when diagnosing skew.

Per-segment counting is side-effect-free: `_count_segment` returns a count (no accumulator), `_print_per_segment_counts` collects and reports.

`--per-segment` uses `default=argparse.SUPPRESS` (like `--XWaitForDPU`) so it's only forwarded when set — a plain `default=False` would cross the argv wire as the string `"False"`, which the server would read as truthy and fire the report unconditionally.

## Docs

README updated in both the examples block and the `scancount` reference section (the other #190 ask).

## Testing

- 16 tests extracted/added (`test_scancount.py`), RED-first then GREEN.
- scancount module at 100% line + branch coverage.
- Full offline suite green (`make test`): 1374 passing.

## Notes

Clean re-submission of #190, isolated from the unrelated poison-pill / bootstrap / CI changes its original stacked branch had bundled in. Only file touched in common with any other open PR is `README.md` (this edits the scancount section; #231 edits the capacity-warnings section) — trivial rebase for whichever merges second.

Per-segment scans not yet run live on Glue — offline-verified only.

### DIFF
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -77,6 +77,15 @@ Here are some example use cases:
 # Use single quotes around the JSON and double quotes within!
 ./bulk scancount --table t --index i --filter-expression "a = :aval" --expression-values '{":aval":"foo"}'
 
+# Add --per-segment to print each segment's item count (sorted, with a
+# skew ratio and a warning when the hottest segment exceeds 5x the mean).
+# Useful for diagnosing hot partitions / uneven key distribution.
+./bulk scancount --table t --per-segment
+
+# By default scancount uses 200 parallel scan segments. Use --segments to
+# tune that — fewer for small tables, more to increase per-segment resolution.
+./bulk scancount --table t --per-segment --segments 50
+
 
 # Compare two tables for differences (uses segmented scans internally)
 ./bulk diff --table t --table2 t2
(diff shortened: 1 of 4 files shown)


hunterhacker:
Lotta duplicated code here. Why copy-paste duplicate the scan loop?

relentlesscol:
Fair hit — that was a copy-paste. Fixed in ca667a4: pulled the shared rate-limited segmented `Select=COUNT` scan into one `_scan_and_count(...on_error)`. `_count_data` and `_count_segment` are now thin wrappers that only inject their distinct failure handling (error_accumulator + worker print vs `log.warning`). The scan loop lives in exactly one place now.

Also added live e2e coverage since `--per-segment` had none — `tests/e2e/connector/test_scancount_smoke.py` runs a real Glue job and asserts the skew report actually renders (segment table + skew ratio), not just that the unit mocks pass. Heads up: I haven't run it against a live account yet (needs creds + bootstrap), so consider that smoke unverified until someone runs `make test-e2e-connector` — I'll do that and report the result rather than leave it claimed-but-unrun. The 63 offline unit tests stay green after the refactor. 🦾

relentlesscol:
Ran it live now (account 654654401288 / us-east-1) — both smokes **pass** against a real Glue job:

- `scancount` → job `SUCCEEDED` (~1:14), `Total records counted: 500`
- `scancount --per-segment` → job `SUCCEEDED` (~1:14), skew report renders: 10 segments, mean 50.0, **skew ratio 1.52x**, all-segment table printed.

Worth flagging what the live run caught that the unit mocks could not: on the pre-merge branch the job crashed at the pricing lookup (`KeyError: 'group'`) *before* reaching any scancount code — the DynamoDB vector products with no `'group' attribute. Merged `main` to pick up #268/#272 (fail-soft pricing) and the smoke went green. That's exactly the false-security gap unit tests leave: 63 green offline tests, real job dead on arrival. Merge is on this branch now; offline suite is 1484 passing after the merge. 🦾

Issue #131, opened 2026-03-29
Allow LiveTail to reconnect if disconnected


hunterhacker:
Long running jobs can drop the LiveTail connection:

```
2026-03-28 21:21:24,936 ERROR [Thread-3 (_watch_log_group)] root - Unexpected error occurred in /aws-glue/jobs/output live tail: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read)).
2026-03-28 21:21:32,207 ERROR [Thread-2 (_watch_log_group)] root - Unexpected error occurred in /aws-glue/jobs/error live tail: AWSHTTPSConnectionPool(host='streaming-logs.us-east-1.amazonaws.com', port=443): Read timed out..
```

In this case, it happened after about 34 minutes. It didn't stop execution, but there was no command output after this point.

Seen several times. Didn't use to see it.

Should catch such an exception and reestablish the tail.

ruskindantra:
This issue can now be resolved

PR #243, opened 2026-07-23
Adding cross-region support for diff

linuswagner:
*Issue #, if available:* n.a.

*Description of changes:* The bulk operation `diff` could only run in a single region because boto3 never gets set up with a region and therefore always infers the region from the environment. Consequently, DDBs in a different region is not found. This PR adds support for tables in different regions for the `diff` operation. If a table is provided as an ARN, the region from the ARN is used. If a table is provided with just its name, the region is still inferred from the environment, making this change backwards compatible. The aggregator is always using the default region since it must run in the region that the tools were bootstrapped in.

*Testing:* Several new unit tests test the new functionality:
`TestCrossRegionDiff`: tests `diff_segment()` and `run()` cross-region behavior
- Two ARNs from different regions create `RateLimiterWorker` instances with the correct per-table region
- Two plain table names both use the default region
- One ARN + one plain name results in the ARN region and default region respectively
- `print_dynamodb_table_info extracts the region` from an ARN and passes it to scan cost calculation
- Plain table name falls back to the default region
- run() creates DynamoDB clients with the correct regions for each table ARN
- `RateLimiterAggregator` is initialized with table1's inferred region

`TestRateLimiterRegionSupport`: tests `RateLimiterWorker` region forwarding:
- `RateLimiterWorker` receives the correct `region_name` kwarg from `infer_region`
- If `infer_region` raises ValueError, `diff_segment` propagates it

`TestInferRegion`: tests `infer_region()` and `_region_from_table_ref()`:
- A DynamoDB ARN returns the region embedded in the ARN
-Various ARNs correctly extract different regions
- A plain table name returns the boto3 session's region
- Falls back to `AWS_REGION` env var when session has no region
- Falls back to `AWS_DEFAULT_REGION` env var
- Raises `ValueError` when no region can be determined
- ARN region wins over whatever the session reports
- A non-DynamoDB ARN (e.g. S3) falls back to the default region

Additionally, I executed the possible variants of cross-account diffs after bootstrapping the executing account (due to not having two identically schemed tables available in a single account, I omitted the case "no table has ARN". Table names and account IDs altered):

```bash
# both tables have ARN
./bulk bootstrap --XRole READ-ONLY # default region set to eu-west-1, account is first table account
./bulk diff --table arn:aws:dynamodb:eu-west-1:123456789012:table/t --table2 arn:aws:dynamodb:eu-south-2:987654321098:table/t2 --format full

# first table has ARN
./bulk bootstrap --XRole READ-ONLY # default region set to eu-south-2, account is second table account
./bulk diff --table arn:aws:dynamodb:eu-west-1:123456789012:table/t --table2 t2 --format full

# second table has ARN
./bulk bootstrap --XRole READ-ONLY # default region set to eu-west-1, account is first table account
./bulk diff --table t --table2 arn:aws:dynamodb:eu-south-2:987654321098:table/t2 --format full
```

All queries execute successfully. Console correctly indicates that the prices from `eu-west-1` and `eu-south-2` are being used.

Also ran tests using `make test`: 1375 passed, 49 skipped, 9 warnings

### DIFF
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -79,7 +79,10 @@ Here are some example use cases:
 
 
 # Compare two tables for differences (uses segmented scans internally)
+# You can specify a name or a full ARN. Using an ARN lets you drive work cross-region and cross-account!
+# If going cross-account, you need a resource-based policy on the table to allow access.
 ./bulk diff --table t --table2 t2
+./bulk diff --table arn:aws:dynamodb:us-east-1:123456789012:table/t --table2 arn:aws:dynamodb:us-west-2:987654321098:table/t2
 
 # The default diff format is "keys" to show primary keys having changes with +/-/* for adds/removes/changes
 # Specifying the format "full" outputs the total items with +/- for before/after
--- a/tools/bulk_executor/README.md
+++ b/tools/bulk_executor/README.md
@@ -470,10 +470,34 @@ If you ever want to stop execution early, you can hit Control-C. The interrupt w
 #### `diff`
 
 * Performs parallel scans to two tables to compare for differences.
-* Requires `table` and `table2` parameters.
+* Requires `table` and `table2` parameters. Parameters can be table names or table ARNs. Using ARNs allows cross-account / cross-region access.
 * Accepts an optional `format` which can be `compact` (prints primary keys of all changed items with `+`, `-`, `*` for adds, removes, changes) or `full` (prints the full values of all changed items with `+` and `-` for before and after). Default is `compact`.
 * Accepts an optional `sample-fraction` to indicate a fraction of the two tables to compare, between 0.0 and 1.0. Can help with sanity checking that runs faster and at lower cost.
 * Accepts an optional `s3` parameter to output the result to S3.
+
+If doing cross-account, you need a resource-based policy to enable access. The following example RBP allows access from two specific roles in the 123456789012 account. The `role/ClientSide` is whatever role you have for the command-line Python program (so the client-side can describe the table and estimate costs). The `role/AWSGlueServiceRoleBulkDynamoDB-DdbReadWrite-us-east-1` is whatever role you have attached to the Glue job (so the `diff` can be performed).
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "CrossAccountIdentityBasedPolicy",
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": [
+          "arn:aws:iam::123456789012:role/ClientSide",
+          "arn:aws:iam::123456789012:role/AWSGlueServiceRoleBulkDynamoDB-DdbReadOnly-us-east-1"
+        ]
+      },
+      "Action": [
+          "dynamodb:DescribeTable",
+          "dynamodb:Scan"
+      ],
+      "Resource": "arn:aws:dynamodb:us-west-1:599882009758:table/table2"
+    }
+  ]
+}
+```
 
 #### `load`
 
--- a/tools/bulk_executor/server/src/python_modules/diff.py
+++ b/tools/bulk_executor/server/src/python_modules/diff.py
@@ -12,7 +12,10 @@ from python_modules.shared.errors import get_error_message
 from python_modules.shared.table_info import (
     get_and_print_dynamodb_table_info,
     get_and_print_table_scan_cost,
-    get_dynamodb_throughput_configs
+    get_dynamodb_throughput_configs,
+    _region_from_table_ref,
+    infer_region,
+    _default_region
 )
 
 from python_modules.shared.rate_limiter import (
@@ -155,13 +158,17 @@ def log_diff(symbol, stream, concise_format):
 
 
 def diff_segment(stream_a_name, stream_b_name, monitor_options_a, monitor_options_b, segment, total_segments, consistent_read, concise_format, job_id, use_s3, bucket, schema_broadcast, rate_limiter_shared_config):
+    table1_region = infer_region(stream_a_name)
+    table2_region = infer_region(stream_b_name)
     rate_limiter_worker_a = RateLimiterWorker(
         shared_config=rate_limiter_shared_config,
+        region_name=table1_region,
         **monitor_options_a
     )
 
     rate_limiter_worker_b = RateLimiterWorker(
         shared_config=rate_limiter_shared_config,
+        region_name=table2_region,
         **monitor_options_b
     )
 
@@ -289,7 +296,7 @@ def diff_segment(stream_a_name, stream_b_name, monitor_options_a, monitor_option
     return diff[0:PRINT_LIMIT]
 
 def print_dynamodb_table_info(table_name, fraction=1.0):
-    region_name = boto3.Session().region_name
+    region_name = _region_from_table_ref(table_name) or _default_region()
     table_info = get_and_print_dynamodb_table_info(table_name)
     return get_and_print_table_scan_cost(table_info, region_name, fraction=fraction)
 
@@ -323,8 +330,10 @@ def run(job, spark_context, glue_context, parsed_args):
     print(f"TOTAL DynamoDB cost for scanning both tables (approx): ${total_cost:,.2f}")
     print()
 
-    schema1 = boto3.client("dynamodb").describe_table(TableName=table1)['Table']['KeySchema']
-    schema2 = boto3.client("dynamodb").describe_table(TableName=table2)['Table']['KeySchema']
+    table1_region = infer_region(table1)
+    table2_region = infer_region(table2)
+    schema1 = boto3.client("dynamodb", region_name=table1_region).describe_table(TableName=table1)['Table']['KeySchema']
+    schema2 = boto3.client("dynamodb", region_name=table2_region).describe_table(TableName=table2)['Table']['KeySchema']
 
     def extract_keys(schema):
         pk = next(e['AttributeName'] for e in schema if e['KeyType'] == 'HASH')
@@ -350,7 +359,8 @@ def run(job, spark_context, glue_context, parsed_args):
         job_run_id=job_id
     )
 
-    rate_limiter_aggregator = RateLimiterAggregator(shared_config=rate_limiter_shared_config)
+    # uses S3 from bootstrapping process -> should use same default region
+    rate_limiter_aggregator = RateLimiterAggregator(shared_config=rate_limiter_shared_config, region_name=_default_region())
 
     monitor_options_1 = get_dynamodb_throughput_configs(parsed_args, table1, modes=("read"), format="monitor")
     monitor_options_2 = get_dynamodb_throughput_configs(parsed_args, table2, modes=("read"), format="monitor")
--- a/tools/bulk_executor/server/src/python_modules/shared/export/pipeline/__init__.py
+++ b/tools/bulk_executor/server/src/python_modules/shared/export/pipeline/__init__.py
@@ -3,7 +3,7 @@ import time
 from ...bulk_executor_error import BulkExecutorError
 from ...logger import log
 from ...errors import ListAccumulator
-from ...table_info import get_dynamodb_throughput_configs
+from ...table_info import get_dynamodb_throughput_configs, infer_region
 from ...rate_limiter import RateLimiterAggregator, RateLimiterSharedConfig
 
 from .validator import validate
@@ -103,11 +103,12 @@ def run_export_pipeline(spark_context, parsed_args, transform_package, post_vali
     table_name = parsed_args.get('table')
     s3_path = parsed_args.get('s3_path')
     transform_name = parsed_args.get('transform')
+    region_name = infer_region(table_name)
 
     bucket_name = parsed_args.get('s3-bucket-name')
     job_run_id = parsed_args.get("JOB_RUN_ID")
     rate_limiter_shared_config = RateLimiterSharedConfig(bucket=bucket_name, job_run_id=job_run_id)
-    rate_limiter_aggregator = RateLimiterAggregator(shared_config=rate_limiter_shared_config)
+    rate_limiter_aggregator = RateLimiterAggregator(shared_config=rate_limiter_shared_config, region_name=region_name)
 
     monitor_options = get_dynamodb_throughput_configs(parsed_args, table_name, modes=["write"], format="monitor")
     log.debug(f"monitor_options {monitor_options}")
--- a/tools/bulk_executor/server/src/python_modules/shared/export/writers/batch_writer.py
+++ b/tools/bulk_executor/server/src/python_modules/shared/export/writers/batch_writer.py
@@ -6,6 +6,7 @@ from botocore.config import Config
 from typing import Dict, Any, Iterator
 from .base_writer import DynamoDBWriter
 from ...rate_limiter import RateLimiterWorker
+from ...table_info import infer_region
 from ...logger import log
 from ...errors import get_error_code, get_error_message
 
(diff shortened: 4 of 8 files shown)


tools/bulk_executor/README.md
linuswagner, line 1: Should also expand the `diff` description further down the file:
- can use ARN for cross-region/account
- add example policy for easier use

linuswagner (commented):
additional doc needed

hunterhacker:
My prompt to Claude:

I have a PR at https://github.com/awslabs/amazon-dynamodb-tools/pull/243/files and I'm wondering why the change adjusted [shared/rate_limiter/__init__.py](https://github.com/awslabs/amazon-dynamodb-tools/pull/243/files#diff-cbf961f3c60244fed9dedb59695dc580dc19dcedd676440193c4fe440e846a40) to do         `self.session = Session(region_name=region_name)` because I don't want the rate limiter running in a different region even if the table ARNs are different regions.

Claude said:

I'll look at the local rate limiter code and the PR diff.

## Why the author did it

`RateLimiterWorker` hands its session to callers via `get_session()`, and every caller builds its DynamoDB client from it with no explicit region:

```python
session = rate_limiter_worker.get_session()
... session.client('dynamodb')   # region comes from the Session
```

It has to be *that* session, because `DynamoDBMonitor` registers the rate-limiting hooks on `session.events` (`before-call.dynamodb.*` etc.). A client built from a different session isn't rate limited at all. So regionalizing the `Session` was the only lever that moved the DDB client without touching all seven call sites.

## But your objection is right — and it's a bug, not a preference

That same session also builds the coordination client:

```python
# DistributedDynamoDBMonitorWorker.py:24
self.s3_client = session.client('s3')
```

So `region_name=` now also relocates the rate limiter's S3 heartbeat/summary traffic to the table's region, while the bucket lives in the bootstrap region — and the PR pins the aggregator to `_default_region()` precisely because of that ("uses S3 from bootstrapping process -> should use same default region"). Worker and aggregator end up disagreeing.

`put_object` against the wrong regional endpoint gets `PermanentRedirect` / `AuthorizationHeaderMalformed`. And it fails *silently* — `_sync_loop` wraps everything in `except Exception as e: print(...)`, so the worker just never publishes metrics and never reads the summary. Rate limiting quietly degrades to per-worker initial rates. That's likely why the author's manual cross-region runs "executed successfully."

## Suggested fix

Keep the regional session for DynamoDB (needed for the event hooks), pin S3 to the default region:

```python
# rate_limiter/__init__.py
def __init__(self, shared_config, region_name=None, **monitor_options):
    self.session = Session(region_name=region_name)      # DDB + monitor hooks
    self.rate_limiter_monitor_worker = DistributedDynamoDBMonitorWorker(
        session=self.session,
        s3_client=Session().client('s3'),                # coordination: bootstrap region
        ...
    )
```

with `DistributedDynamoDBMonitorWorker(..., s3_client=None)` defaulting to `session.client('s3')` for backward compatibility. Same treatment for the aggregator — it only ever uses S3, so `region_name` there is misleading at best.

## One more thing to flag in review

`shared/export/pipeline/__init__.py` does the opposite of what the PR description claims:

```python
region_name = infer_region(table_name)
rate_limiter_aggregator = RateLimiterAggregator(shared_config=..., region_name=region_name)
```

That points the aggregator's S3 client at the *table's* region, contradicting `diff.py`'s `_default_region()`. Same root cause, more directly wrong.

Sources: [[PR #243 files](https://github.com/awslabs/amazon-dynamodb-tools/pull/243/files)](https://github.com/awslabs/amazon-dynamodb-tools/pull/243/files)

And I asked:

How did I solve this for the copy code?

Claude:

You already solved it the right way in `copy.py` — at the **client**, not the Session:

```python
source_session = source_rl.get_session()          # plain Session(), default region
...
# Talk to the right region if the table name is an ARN to a diff region
source_region = _region_from_table_ref(source_table) or source_session.region_name
target_region = _region_from_table_ref(target_table) or target_session.region_name

source_ddb = source_session.resource("dynamodb", config=cfg, region_name=source_region)
target_ddb = target_session.resource("dynamodb", config=cfg, region_name=target_region)
```

`RateLimiterWorker` / `RateLimiterAggregator` are constructed with no `region_name` at all, so their S3 coordination traffic stays in the bootstrap region. The per-table region rides on the DynamoDB resource only. And because that resource still comes from the rate limiter's session, the `session.events` hooks continue to apply — you get cross-region *and* rate limiting.

So the review comment on PR #243 is basically "do what `copy.py` does": leave `RateLimiterWorker` alone, and in `diff.py` pass `region_name=` into the DynamoDB client that `SegmentStream` builds. Right now `SegmentStream` takes a session and presumably calls `session.client("dynamodb")` internally — that's the one spot that needs the region threaded through.

One difference worth deciding: `copy.py` uses `_region_from_table_ref(x) or session.region_name` (silently proceeds with `None` if neither resolves), while the PR's `infer_region()` raises `ValueError`. The PR's version is stricter and probably better — if you take it, `copy.py` should switch to it too so there's one rule.

hunterhacker:
Reviewing the code (haven't tried running it yet) it looks good. 

I notice the use of 599882009758 in the README, which should be a non-realistic account id. 

The `diff` samples under "Usage samples" could include a call using ARNs as well, similar to how `copy` does it.

linuswagner:
> Reviewing the code (haven't tried running it yet) it looks good.

Me neither, I'll do this now

> I notice the use of 599882009758 in the README, which should be a non-realistic account id.

Corrected, also in the existing policy for `copy`

> The `diff` samples under "Usage samples" could include a call using ARNs as well, similar to how `copy` does it.

Isn't that already there? What isn't there for both `copy` and `diff` is an example where just one of the two is an ARN

linuswagner:
Executing the first manual test lead to
```sh
./bulk diff --table arn:aws:dynamodb:eu-south-2:XXXXXX:table/t --table2 arn:aws:dynamodb:us-east-1:XXXXXX:table/t2 --format full

    Bulk Executor environment configurations:

      AWS -
        Account: XXXXXX
        Region: eu-south-2
        Role: arn:aws:sts::XXXXXX:assumed-role/XXXXXX
    
Running action 'diff' with arguments: {'verb': 'diff', 'table': 'arn:aws:dynamodb:eu-south-2:XXXXXX:table/t', 'table2': 'arn:aws:dynamodb:us-east-1:XXXXXX:table/t2', 's3': False, 'sample_fraction': 1.0, 'format': 'full'}


The bulk executor job cost consists of DynamoDB and Glue costs
For small jobs, the Glue cost portion is usually dominating
Using fewer Glue workers for small jobs, through the --XNumberOfWorkers parameter, will often reduce the Glue costs
For large jobs, where the cost is more significant, the DynamoDB cost portion is usually dominating
The DynamoDB cost will be estimated below
The Glue cost estimation isn't provided since it is based on DPU hours being used by the job, which is hard to estimate in advance
You can run the script with the --XWaitForDPU parameter in order to print the used Glue DPU hours at the end of the job

Starting Bulk Executor Glue Job...
Bulk Executor Glue Job started with job run ID: XXXXXX
Job start time: 2026-08-13 11:44:54.387593
Press Ctrl+C to cancel Bulk Executor Glue Job.
2026-08-13 11:46:16,794 ERROR - Unexpected error occurred in /aws-glue/jobs/error live tail: AWSHTTPSConnectionPool(host='stream-logs.eu-south-2.amazonaws.com', port=443): Read timed out..
2026-08-13 11:46:16,919 ERROR - Unexpected error occurred in /aws-glue/jobs/output live tail: AWSHTTPSConnectionPool(host='stream-logs.eu-south-2.amazonaws.com', port=443): Read timed out..
Job completed successfully. Job duration: 0:01:53
```

Redriving the actual job leads to a successful execution.

The same happened for the 3rd test.
2nd execution leads to only one timeout:
```
2026-08-13 12:06:24,154 ERROR - Unexpected error occurred in /aws-glue/jobs/output live tail: AWSHTTPSConnectionPool(host='stream-logs.us-east-1.amazonaws.com', port=443): Read timed out..
```
3rd execution does not show difference:
```
...
TOTAL DynamoDB cost for scanning both tables (approx): $0.00

[arn:aws:dynamodb:eu-south-2:XXXXXX:table/XXXXXX] Max read rate set to account quota limit: 40000
[XXXXXX] Max read rate set to account quota limit: 40000
Job completed successfully. Job duration: 0:01:54
```
The 4th execution showed the expected differences.

The 2nd test worked right away.

---
Initial investigation:
- The tables I am testing on are rather small (10 items)
- The default split of 400 leads to a lot of executors having no real work
- By default, 220 workers are present, again a lot of 10 items
- Initial PR involved one table that was significantly larger -> maybe not an issue there
- ran 2nd test on commit 3cbe8ff888c07c0eb217d56cfe0ffe3e7e84da7c (before this change) checking the same table in the same account (i.e. `t == t2`). The output differs between execution 2 and 3 (1 and 2 are identical) and parts of the output are missing:

exec 2
```
./bulk diff --table XXXXXX --table2 XXXXXX --format full

    Bulk Executor environment configurations:

      AWS -
        Account: XXXXXX
        Region: us-east-1
        Role: arn:aws:sts::XXXXXX:assumed-role/XXXXXX
    
Running action 'diff' with arguments: {'verb': 'diff', 'table': 'XXXXXX', 'table2': 'XXXXXX', 's3': False, 'sample_fraction': 1.0, 'format': 'full'}


The bulk executor job cost consists of DynamoDB and Glue costs
For small jobs, the Glue cost portion is usually dominating
Using fewer Glue workers for small jobs, through the --XNumberOfWorkers parameter, will often reduce the Glue costs
For large jobs, where the cost is more significant, the DynamoDB cost portion is usually dominating
The DynamoDB cost will be estimated below
The Glue cost estimation isn't provided since it is based on DPU hours being used by the job, which is hard to estimate in advance
You can run the script with the --XWaitForDPU parameter in order to print the used Glue DPU hours at the end of the job

Starting Bulk Executor Glue Job...
Bulk Executor Glue Job started with job run ID: XXXXXX
Job start time: 2026-08-13 12:32:44.697428
Press Ctrl+C to cancel Bulk Executor Glue Job.
Table: 'XXXXXX'
Billing mode: On-demand

Table Item Count (approx): 10
Table Size (approx): 1,279 bytes

DynamoDB read costs depend on the table size.
DynamoDB read units required for a full scan (approx): 1
Approx DynamoDB cost for on-demand scan consuming 1 RRUs (using us-east-1 prices): $0.00


Table: 'XXXXXX'
Billing mode: On-demand

Table Item Count (approx): 10
Table Size (approx): 1,279 bytes

DynamoDB read costs depend on the table size.
DynamoDB read units required for a full scan (approx): 1
Approx DynamoDB cost for on-demand scan consuming 1 RRUs (using us-east-1 prices): $0.00


TOTAL DynamoDB cost for scanning both tables (approx): $0.00

[XXXXXX] Max read rate set to account quota limit: 40000
[XXXXXX] Max read rate set to account quota limit: 40000
No differences found

Job completed successfully. Job duration: 0:02:24
```

exec 3
```
./bulk diff --table XXXXXX --table2 XXXXXX --format full

    Bulk Executor environment configurations:

      AWS -
        Account: XXXXXX
        Region: us-east-1
        Role: arn:aws:sts::XXXXXX:assumed-role/XXXXXX
    
Running action 'diff' with arguments: {'verb': 'diff', 'table': 'XXXXXX', 'table2': 'XXXXXX', 's3': False, 'sample_fraction': 1.0, 'format': 'full'}


The bulk executor job cost consists of DynamoDB and Glue costs
For small jobs, the Glue cost portion is usually dominating
Using fewer Glue workers for small jobs, through the --XNumberOfWorkers parameter, will often reduce the Glue costs
For large jobs, where the cost is more significant, the DynamoDB cost portion is usually dominating
The DynamoDB cost will be estimated below
The Glue cost estimation isn't provided since it is based on DPU hours being used by the job, which is hard to estimate in advance
You can run the script with the --XWaitForDPU parameter in order to print the used Glue DPU hours at the end of the job

Starting Bulk Executor Glue Job...
Bulk Executor Glue Job started with job run ID: XXXXXX
Job start time: 2026-08-13 12:35:14.436288
Press Ctrl+C to cancel Bulk Executor Glue Job.
No differences found

Job completed successfully. Job duration: 0:01:39
```

## Conclusion
- At least on small tables, maybe even on bigger ones, the current code has trouble to produce the correct output.
- It seems like the cross-region feature amplifies this problem as failures were more frequent and obvious (errors instead of missing messages)

linuswagner:
I poked AI for this issue (didn't have time to check for plausibility yet), here's the analysis:

---

## Live tail silently drops output when CloudWatch Logs streaming connection times out

### Problem

When a Glue job runs, the client streams its output in real-time using the CloudWatch Logs `start_live_tail` API. If the HTTP/2 streaming connection to the `stream-logs.<region>.amazonaws.com` endpoint times out (read timeout), the log-watching thread exits permanently and all subsequent job output is lost on the client side.

The user sees partial output (e.g. table cost info printed early in the job) followed by "Job completed successfully", but misses the actual results (e.g. diff output, counts) that were printed later in the job's execution.

### Root Cause

In `client/src/runner.py`, the `_watch_log_group` method handles connection failures:

```python
except (ConnectionError, HTTPClientError, EventStreamError) as e:
    # Reconnects on these transient errors ✓
    ...
except Exception as e:
    log.error(f"Unexpected error occurred in {log_group_name} live tail: {str(e)}.")
    return  # ← Exits the thread permanently, no reconnection attempt
```

A `ReadTimeoutError` (from urllib3/botocore) does not inherit from `ConnectionError`, `HTTPClientError`, or `EventStreamError`. It falls into the generic `Exception` handler which logs an error and **permanently exits** the thread — no reconnection is attempted.

### Impact

- Job output is silently lost on the client side (the Glue job itself succeeds and its output is in CloudWatch Logs)
- Users may believe the job produced no output when it actually did
- Particularly affects long-running jobs or jobs connecting to remote-region CloudWatch endpoints (higher latency = more timeouts)

### Observed Error

```
ERROR - Unexpected error occurred in /aws-glue/jobs/output live tail: AWSHTTPSConnectionPool(host='stream-logs.eu-south-2.amazonaws.com', port=443): Read timed out..
```

### Proposed Fix

Treat `ReadTimeoutError` as a transient/recoverable error and reconnect, similar to how `ConnectionError` is already handled:

```python
except (ConnectionError, HTTPClientError, EventStreamError) as e:
    # existing reconnection logic
    ...
except Exception as e:
    if 'timed out' in str(e).lower():
        # Treat as transient — check if job is done, otherwise reconnect
        job_run_state = self._get_job_run_state(job_run_id)
        if job_run_state in TERMINAL_JOB_STATES or job_run_state == SUCCEEDED_STATE or job_unhealthy_event.is_set():
            return
        log.debug(f"Live tail for {log_group_name} timed out, reconnecting...")
        time.sleep(1)
    else:
        log.error(f"Unexpected error occurred in {log_group_name} live tail: {str(e)}.")
        return
```

Alternatively, import `ReadTimeoutError` from botocore and add it to the first `except` clause.

### Additional consideration

Even with reconnection, there may be a gap in log delivery between the timeout and reconnection. Consider adding a fallback that uses `get_log_events` / `filter_log_events` after the job completes to retrieve any logs that were missed during live tail gaps.

---

hunterhacker:
Just a note, we faced this kind of issue previously: https://github.com/awslabs/amazon-dynamodb-tools/issues/131

hunterhacker:
Thanks for this, @linuswagner! Since this was opened, `main` moved on and #282 rewrote `diff.py`'s output path (diff now always writes the full result to S3 and prints a bounded console preview), so this branch no longer rebases cleanly onto it.

I've opened #283, which re-applies your cross-region approach on top of current `main` and keeps you credited as co-author. It's scoped to `diff` only — I left out the `batch_writer.py` (load write-path) change, since that was undocumented/untested here and is better as its own deliberate PR. It also tightens the cross-account RBP example to read-only (`DescribeTable` + `Scan`) since `diff` never writes.

Superseded by #283.

hunterhacker:
Integrated key parts into #283, now merged.

